PR #11, opened 2021-09-09
feat: add negotiation support for go-extensions

jaredallard:
**What this PR does / why we need it**: This PR adds support for basic go extensions to stencil. An extension is made via go-plugin. A basic example of a plugin looks like so:

```go
type TestPlugin struct{}

func (tp *TestPlugin) GetConfig() (*apiv1.Config, error) {
	return &apiv1.Config{
		Name: "test",
	}, nil
}

func (tp *TestPlugin) ExecuteTemplateFunction(t *apiv1.TemplateFunctionExec) (interface{}, error) {
	return hello()
}

func (tp *TestPlugin) GetTemplateFunctions() ([]*apiv1.TemplateFunction, error) {
	return []*apiv1.TemplateFunction{
		{
			Name:          "hello",
			ArgumentTypes: []interface{}{""},
			ReturnType:    "",
		},
	}, nil
}

func hello() (interface{}, error) {
	return "hello from a plugin!", nil
}

func main() {
	err := apiv1.NewExtensionImplementation(&TestPlugin{})
	if err != nil {
		logrus.WithError(err).Fatal("failed to start extension")
	}
}

```

When run via stencil, with a module manifest like so:

```yaml
name: test-plugin
type: extension
```

the extension will be executed by stencil and automatically exposed to templates. Templates can call a plugin via the `extensions.Call` function. For example, to call that test function would be done like so from a template:

```tpl
{{- extensions.Call "test.hello" }}

# results in: hello from a plugin!
```

This is done by dynamically generating a function via `reflect.MakeFunc` by communicating with the plugin in the extensions package in this PR based on the `GetTemplateFunctions` RPC. When this function is called via `extensions.Call` in the template, a go-plugin is transparently called over RPC and the response is returned. We're limited to only types that gob supports, but in general, this is pretty powerful.

**JIRA ID**: [DTSS-765]

**Notes for your reviewer**:

[DTSS-765]: https://outreach-io.atlassian.net/browse/DTSS-765

### DIFF
--- a/pkg/extensions/apiv1/extensions.go
+++ b/pkg/extensions/apiv1/extensions.go
@@ -1,0 +1,74 @@
+// Package apiv1 implements the extension API for stencil
+package apiv1
+
+import (
+	"github.com/getoutreach/stencil/pkg/functions"
+)
+
+const (
+	// Version that this extension API implements
+	Version = 1
+
+	// CookieKey is a basic UX feature for ensuring that
+	// we execute a valid stencil plugin. This is exported
+	// for ease of consumption by extensions.
+	CookieKey = "STENCIL_PLUGIN"
+
+	// CookieValue is the expected value for our CookieKey to
+	// return.
+	CookieValue = "はじめまして"
+)
+
+// TemplateFunction is a request to create a new template function.
+type TemplateFunction struct {
+	// Name of the template function, will be registered as:
+	//  extensions.<extensionLowerName>.<name>
+	Name string
+
+	// NumArguments are the number of arguments, arbitrarily, that this
+	// function expects.
+	NumArguments int
+}
+
+// TemplateFunctionExec executes a template function
+type TemplateFunctionExec struct {
+	// Name is the name of this go-template function. It will be prefixed with the
+	// following format: extensions.<name>.<templateName>
+	Name string
+
+	// File is the file metadata provided by stencil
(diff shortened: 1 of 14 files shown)


jaredallard:
Still a WIP, but in case you have thoughts @vk-outreach :)

pkg/extensions/apiv1/extensions.go
```diff
@@ -0,0 +1,74 @@
...
+	// GetTemplateFunctions returns all go-template functions this ext
+	// implements, when a function is called, it's transparently passed over to
+	// the actual extension and called there instead, it's output being
+	// returned.
+	GetTemplateFunctions() ([]*TemplateFunction, error)
+
+	// ExecuteTemplateFunction executes a provided template function
+	// and returns it's response.
+	ExecuteTemplateFunction(t *TemplateFunctionExec) ([]interface{}, error)
+}
```
vk-outreach, line 74: Hmm.. I suggest considering a different route to installing plugin methods: let stencil expose a grpc server that plugins can use to call into stencil for any things they need to get done (including installing functions).

Ideally, stencil will have a global table of (key, value) entries with grpc methods to CRUD this table.    This table can just be exposed directly to all templates (including crud methods on it).

jaredallard: It didn't look like that worked with `go-plugin`, but I can certainly re-evaluate. Their examples are the most obvious.

jaredallard:
We're looking to icebox this for a bit now due to some internal direction change, so I want to leave this at its current implementation for now. Please take a look @vk-outreach!

PR #42, opened 2021-12-29
feat: update with bootstrap, run bootstrap

jaredallard:

This PR runs bootstrap on stencil and updates it to match close to platinum lintroller as well as syncs the processor framework changes in Bootstrap, but removes the `go` and `gobox` processors as those shouldn't be in stencil long-term.

I've also noted in the code that the processor framework should be deprecated. It's not built to handle arbitrary repositories and shouldn't be built into stencil like it currently is now (baked in processors, that is).

**JIRA ID**: [DTSS-0]


deregtd (approved):
tsk tsk, so much rework inside of the re-bootstrap commit

jaredallard:
> tsk tsk, so much rework inside of the re-bootstrap commit

who are u

PR #43, opened 2022-01-08
fix: work with bootstrap

jaredallard:

This PR does a bunch of locally tested of fixes that were required to get this to work seamlessly with Bootstrap. I also focused on adding a lot more code documentation.

**JIRA ID**: [DTSS-1138]

## Notes for your reviewers

[DTSS-1138]: https://outreach-io.atlassian.net/browse/DTSS-1138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/stencil/stencil.go
```diff
@@ -55,6 +56,11 @@ func main() {
 		Action: func(c *cli.Context) error {
 			log.Infof("stencil %s", oapp.Version)
 
+			if c.Bool("debug") {
+				log.SetLevel(logrus.DebugLevel)
+				log.Debug("Debug logging enabled")
+			}
```
jaredallard, line 62: This is probably something we should be doing by default in bootstrap, will add this later.

cmd/stencil/stencil.go
```diff
@@ -95,11 +108,6 @@ func main() {
 			Name:  "dev",
 			Usage: "Use local manifests instead of remote ones, useful for development",
 		},
-		&cli.StringFlag{
-			Name:    "github-ssh-key",
-			Usage:   "SSH Key to use to download templates with, if not set ~/.ssh/config is read and falls back to ssh-agent",
-			EnvVars: []string{"GITHUB_SSH_KEY"},
-		},
```
jaredallard, line 102: SSH support didn't make much sense when extensions themselves _have_ to be downloaded from Github because of the usage of the Github API. So, it's gone now, and we use the GH cli to dish out the token if not supplied.

pkg/codegen/codegen.go
```diff
@@ -281,47 +322,17 @@ func (b *Builder) makeTemplateParameters(ctx context.Context) (map[string]interf
...
 
 // FetchTemplate fetches a template from a git repository
 func (b *Builder) FetchTemplate(ctx context.Context, filePath string) (io.Reader, error) {
-	f, err := b.GitRepoFs.Open(filePath)
+	f, err := b.templateFS.Open(filePath)
 	return f, errors.Wrap(err, filePath)
 }
 
-// HasDeviations looks for deviation blocks in a file, returning true if they exist
-func (b *Builder) HasDeviations(_ context.Context, filePath string) bool {
```
jaredallard, line 296: Deviations support is a big ol copy of the command parser and, honestly, I'd like to see us not support it in stencil proper. Maybe I'm idealistic here. If there's push back I'm happy to add it back (I'll cleanup the command parser so it's not duplicated at least)

pkg/codegen/codegen.go
```diff
@@ -400,16 +429,15 @@ func (b *Builder) WriteTemplate(ctx context.Context, filePath,
...
-
 	templates, err := b.renderTemplate(filePath, contents, args)
 	if err != nil {
 		return nil, err
 	}
 
+	// TODO(jaredallard): I think this implementation will break if
+	// a multi-file template has blocks with different contents. Need
+	// to look into this. Possible we'll need to pre-render to
+	// determine which files we're writing to to populate args.
```
jaredallard, line 440: This TODO is very valid, thinking through this. It's going to cause issues with any files that generate more files, e.g. the command processor.

We'll need to bring the actual "get blocks" code into this loop as well. This is slightly problematic because we don't know what files we're going to write to until _after_ we execute the template. I'm going to need to think about this more. The naive solution would be to execute it once, read the files, then you'd know, but that probably still won't solve it because the template itself is the one iterating over the files and writing the contents,

Instead, I think the solution is to have `stencil.ApplyTemplate` do the blocks logic. This will require a signature change to have it know which file it's going to be written to (right now it's piped to `stencil.InstallFile`), but this would allow us to support that case.

For context about what I'm going on about/a sample file, see the pkg.go.dev docs: https://pkg.go.dev/github.com/getoutreach/stencil/pkg/functions#Stencil.InstallFile

pkg/codegen/codegen.go
george-e-shaw-iv: this is optimistic

pkg/codegen/codegen.go
```diff
@@ -351,17 +366,30 @@ func (b *Builder) WriteTemplate(ctx context.Context, filePath,
...
+				log := b.log.WithField("command.name", cmd)
+				log.Debug("Processing command")
+
 				switch cmd {
 				case "Block":
 					blockName := matches[3]
 
+					log.WithFields(logrus.Fields{
+						"block.name":  blockName,
+						"block.start": filePath + ":" + strconv.Itoa(i),
```
george-e-shaw-iv, line 378:
```suggestion
						"block.start": fmt.Sprintf("%s:%s", filePath, strconv.Itoa(i)),
```
string concatenation is ugly

pkg/codegen/codegen.go
```diff
@@ -351,17 +366,30 @@ func (b *Builder) WriteTemplate(ctx context.Context, filePath,
...
 					if curBlockName != "" {
 						return nil, fmt.Errorf("invalid Block when already inside of a block, at %s:%d", filePath, i)
 					}
 					curBlockName = blockName
 				case "EndBlock":
 					blockName := matches[3]
 
+					log.WithFields(logrus.Fields{
+						"block.name":  blockName,
+						"block.start": filePath + ":" + strconv.Itoa(i),
```
george-e-shaw-iv, line 390:
```suggestion
						"block.start": fmt.Sprintf("%s:%s", filePath, i),
```

pkg/codegen/codegen.go
jaredallard: ur ugly

jaredallard:
```suggestion
						"block.start": fmt.Sprintf("%s:%d", filePath, i),
```

pkg/codegen/codegen.go
jaredallard:
```suggestion
						"block.start": fmt.Sprintf("%s:%d", filePath, i),
```

george-e-shaw-iv: that what i meant

pkg/gitauth/gitauth.go
```diff
@@ -81,12 +54,13 @@ func configureAccessTokenAuth(token cfg.SecretData, opts *git.CloneOptions) erro
...
-// Github repository clones.
-func ConfigureAuth(sshKeyPath string, accessToken cfg.SecretData, opts *git.CloneOptions, log logrus.FieldLogger) error {
-	if accessToken != "" {
-		return configureAccessTokenAuth(accessToken, opts)
+// Github repository clones
+func ConfigureAuth(accessToken cfg.SecretData, opts *git.CloneOptions,
+	log logrus.FieldLogger) error {
+	// Don't setup auth if no auth token is set
+	if accessToken == "" {
+		return nil
```
george-e-shaw-iv, line 62: could be a good idea to have an info-level log here

george-e-shaw-iv (approved):
looks fine. we should really cut some tickets to remove and rework the entirety of the testing framework. there is 100% a way to make some serious optimizations there and it's completely necessary IMO

pkg/codegen/codegen.go
jaredallard: Will address when we get to it and redesign to be more visible

PR #61, opened 2022-03-14
feat(stencil): Redesign codegen and module system

jaredallard:

This PR introduces a plethora of changes, namely focused on a redesign of the codegen system to have a module and template context attached to all areas of the codegen process. This allows us to better debug where errors are coming from and handle things like file conflicts, namely, among other things.

The main concept introduced in this PR is a heiarchy, that starts with a "module":

Module -> Template -> File. Each tier can contain n number of items below it, as templates can create files, and modules have templates. Keep that in mind when reviewing the PR.

The former package `codegen` has been split into two distinct packages:

 * `cmd/stencil` - a thin wrapper around the new codegen package that writes files to disk and makes "stencil" the stencil command.
 * `codegen` - the core of the codegen system, split out into multiple files with test coverage over the majority of the package. This handles passing functions to templates and creating a potentially future usable go SDK for running stencil in embedded scenarios.

## Jira ID

[DTSS-1376] 
[DTSS-1377] 
[DTSS-1380]

## Notes for your reviewers

[DTSS-1376]: https://outreach-io.atlassian.net/browse/DTSS-1376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/stencil/describe.go
```diff
@@ -0,0 +1,38 @@
...
+
+			fileName := c.Args().First()
+			for _, f := range l.Files {
+				if f.Name == fileName {
+					fmt.Printf("%s was created by module https://%s (template: %s)\n", f.Name, f.Module, f.Template)
+					return nil
+				}
+			}
+
+			return fmt.Errorf("file isn't created by stencil")
```
george-e-shaw-iv, line 35: i understand why this is an error i think (this in theory should never fire), but should you add the filepath to the error

internal/codegen/codegen.go
```diff
@@ -0,0 +1,261 @@
...
+	return nil
+}
+
+// setDefaultArguments translates a few manifest values
+// into arguments that can be accessed via stencil.Arg
+func (b *Builder) setDefaultArguments() error {
+	if b.manifest.Arguments == nil {
+		b.manifest.Arguments = make(map[string]interface{})
+	}
+	b.manifest.Arguments["name"] = b.manifest.Name
```
george-e-shaw-iv, line 135: should these default arguments be underneath a namespace to avoid collision w/other modules?

internal/modules/module.go
```diff
@@ -0,0 +1,204 @@
...
+
+	var manifest configuration.TemplateRepositoryManifest
+	if err := yaml.NewDecoder(mf).Decode(&manifest); err != nil {
+		return configuration.TemplateRepositoryManifest{}, err
+	}
+
+	// ensure that the manifest name is equal to the import path
+	if manifest.Name != m.Name {
+		return configuration.TemplateRepositoryManifest{}, fmt.Errorf(
+			"Module declares it's import path as %q but was imported as %q",
```
george-e-shaw-iv, line 143:
```suggestion
			"module declares it's import path as %q but was imported as %q",
```

internal/modules/modules.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/getoutreach/stencil/pkg/configuration"
+	"github.com/pkg/errors"
+	giturls "github.com/whilp/git-urls"
+)
+
+// GetModulesForService returns a list of modules that a given service manifest
+// depends on. They are not returned in the order of their import.
+func GetModulesForService(ctx context.Context, m *configuration.ServiceManifest) ([]*Module, error) {
+	// We only support importing a single version of a module
+	// IDEA(jaredallard): Always use the latest?
```
george-e-shaw-iv, line 24: we should definitely pin versions - we talked about this a bit in standup the other day

cmd/stencil/describe.go
```diff
@@ -0,0 +1,38 @@
+// Copyright 2022 Outreach Corporation. All Rights Reserved.
+
+// Description: This file contains code for the describe command
+
+package main
+
+import (
+	"fmt"
+
+	pubstencil "github.com/getoutreach/stencil/pkg/stencil"
```
vk-outreach, line 10: why is this `pubstencil` ?

internal/tplfuncs/file.go
```diff
@@ -0,0 +1,103 @@
...
+//   {{- end }}
+func (f *File) Create(path string, mode os.FileMode, modTime time.Time) error {
+	var err error
+	f.f, err = functions.NewFile(path, mode, modTime)
+	if err != nil {
+		return err
+	}
+
+	// If we have a single file with zero contents, replace it
+	if len(f.t.Files) == 1 && len(f.t.Files[0].Bytes()) == 0 {
```
vk-outreach, line 96: This took me a while to grok.  I get that you want to handle the simple case of a template which just renders itself without requiring a `file.Create` on the current template -- and so you defaulted the initial files list in the template constructor to include the current template's equivalent.

But we now have two places where we have this exception (including [here](https://github.com/getoutreach/stencil/pull/61/files#diff-da62329dd65f6ab1c9f34f345743c4a84f57bdfdc905248ff9329e102bdd002bR72))..  I suggest a slightly different pattern:  let `Files` always be just the list of files explicitly created.  Instead, add a helper method `FilesWithDefault()` that both `ApplyTemplate` and `Render` can call.  This function can check if `len(Files) == 0` and create the empty file as needed.

internal/tplfuncs/file.go
```diff
@@ -0,0 +1,103 @@
...
+//   func main() {
+//     fmt.Println("hello, world!")
+//   }
+//
+//   {{- end }}
+//
+//   # Generate a "<commandName>.go" file for each command in .arguments.commands
+//   {{- range $_, $commandName := (stencil.Arg "commands") }}
+//   {{- file.Create (printf "cmd/%s.go" $commandName) 0600 now }}
+//   {{- stencil.ApplyTemplate "commands" | file.SetContents }}
```
vk-outreach, line 86:
```suggestion
//   {{- stencil.ApplyTemplate "command" | file.SetContents }}
```

internal/tplfuncs/file.go
jaredallard: Yeah, not a bad idea. I didn't like the logic here anyways, will fix!

internal/codegen/codegen.go
jaredallard: Hmmmm, yeah I think so. I'll put this under "app"

internal/modules/modules.go
jaredallard: Yeah, this comment was to myself to remove before merging. That's just what to do when a dep is required by the same package @ different versions

internal/tplfuncs/stencil.go
```diff
@@ -0,0 +1,75 @@
...
+	s *functions.Stencil
+	m *configuration.ServiceManifest
+}
+
+// Arg returns the value of an argument in the service's
+// manifest.
+// Note: Only the top-level arguments are supported.
+//
+//   {{- stencil.Arg "name" }}
+func (s *Stencil) Arg(path string) interface{} {
```
vk-outreach: Just FYI -- no action needed.

In my non-outreach version of bootstrap, I have the concept of `prompt args` -- `{{ stencil.Arg "foo" "Do you want foo?" }}` which allows bootstrap to prompt the user to fill out `foo` if it is not present.  In my case, I store this in the lockfile though we could very well store this in the main service.yaml itself...

(In my version, the `prompt` calls will fail in CI if the arg is not present; it won't prompt in CI)

In retrospect, I wish I had some sort of a schema for modules that would make it easy for `stencil describe module-foo` to spit out the arguments and their individual documentation..  So, it might be worth considering if we can add a `stencil.DocumentArg(...)` where the documentation is persisted in the lock file and used in stencil help.

jaredallard: We actually store the docs in the `manifest.yaml` that exists in the module repository, we can totally expose the information in `stencil describe` though. The idea was if something is "required" we'd prompt for a value before rendering the templates -- of course, that's not actually done yet 😓

CONTRIBUTING.md
```diff
@@ -1,25 +1,25 @@
 
-# stencil
+# <no value>
```
george-e-shaw-iv, line 2: 🤔

CONTRIBUTING.md
```diff
@@ -1,25 +1,25 @@
...
+<!--- Block(customBuildingAndTesting) -->
+<!--- EndBlock(customBuildingAndTesting) -->
 ### Replacing a Remote Version of the Library with Local Version
 
 If you want to test the library exposed in this repository in a project that uses it, you can
 add the following `replace` directive to that project's `go.mod` file:
 
 ```
-replace github.com/getoutreach/stencil => /path/to/local/version/stencil
+replace github.com/getoutreach/<no value> => /path/to/local/version/<no value>
```
george-e-shaw-iv, line 22: PANIK

internal/functions/stencil.go
```diff
@@ -0,0 +1,178 @@
...
+			"Generator": app.Info().Name,
+			"Version":   app.Info().Version,
+		},
+		"App": map[string]string{
+			"Name": s.m.Name,
+		},
+		"Repository": map[string]string{
+			"HeadBranch": "main",
+		},
+		"Box": boxConf,
```
george-e-shaw-iv, line 129: what is the behavior if i, as a module, decide one of my argument keys should be named `App`? does this default get overwritten?

internal/functions/tpl_file.go
```diff
@@ -0,0 +1,108 @@
...
+// with the extension '.tpl' removed from the path and operated on.
+type TplFile struct {
+	// f is the current file we're writing to
+	f *File
+
+	// t is the current template
+	t *Template
+}
+
+func (f *TplFile) Spew() string {
```
george-e-shaw-iv, line 28:
```suggestion
// Spew pretty prints the underlying data structure supporting the file for debugging.
func (f *TplFile) Spew() string {
```

internal/functions/stencil.go
vk-outreach: I think arguments can only be accessed via `stencil.Args` ?  Hopefully we are not injecting arguments directly.

It would be nice if all of these were strongly typed (i.e. no `map[string]string{...}`)...

Also, I think `Metadata/App/Repository` don't seem like a clear distinction.. can you summarize what goes where?

Also, I think `metadata/repo/app` seems less descriptive... 

I tend to think in terms of things that are properties of the installation (on disk and not related directly to service.yaml/stencil), things that are a property of the runtime (i.e. stencil host environment + invocation, possibly golang version if we ever need it but also stuff like standard copyright notices) and things that are stencil config (but not specific to a module, things like app name that go into service.yaml):

* install (anything about the current installation: repo name, branch, tag etc).
* runtime (anything about the current runtime environment: calling app name, calling app version, whether running in CI or local etc; access to runtime flags etc could happen via this as well such as debug logging)
* config (any global shared service.yaml config such as app name)

Alternative break downs are ok -- but would prefer if we clearly pinned down the boundaries up-front.

jaredallard: no, these are different. Arguments are accessed via `stencil.Arg(name)` or `stencil.Args().path`, while these come from `.App` or `.Metadata`, so on.

internal/functions/tpl_file.go
jaredallard: i actually meant to delete this :D

internal/functions/stencil.go
george-e-shaw-iv: Makes sense, thats my own misunderstanding then

jaredallard: Ah -- didn't see your comment @vk-outreach (thanks Github). Totally agree, already making them strongly typed and with your distinctions. A much better way of doing what I wanted to do :D

internal/tplfuncs/file.go
jaredallard: So, I played around with that and I don't really like needing to call that logic every time we add a new file to the "stack". Instead, I changed it to be all self-contained in `Render()`. LMK if you like this, if you're not a fan I'm not super attached and can easily go the route you suggested!

internal/functions/stencil.go
jaredallard: Changed this up a little bit... PTAL! (`internal/codegen/values.go`)

vk-outreach (approved):
LGTM ... would be nice to have more tests for the exported packages

PR #62, opened 2022-03-18
feat(codegen): support templates injecting data cross-modules

jaredallard:

This PR introduces the concept of "module hooks" which are cross-module blocks that modules can write to. These enable modules to programmatically insert contents into another module in a predictable way.

There are currently restrictions on the type of data that is allowed to be passed, they must be slices or maps. This helps ensure that multiple modules are able to write to the same hook without the possibility of being overwritten. If the same hook is written more than once, the data is merged together via mergo.

## Jira ID

[DTSS-1379]

## Notes for your reviewers

[DTSS-1379]: https://outreach-io.atlassian.net/browse/DTSS-1379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/codegen/stencil.go
+++ b/internal/codegen/stencil.go
@@ -23,7 +23,7 @@ import (
 
 // NewStencil creates a new, fully initialized Stencil renderer function
 func NewStencil(m *configuration.ServiceManifest, mods []*modules.Module) *Stencil {
-	return &Stencil{m, extensions.NewHost(), mods}
+	return &Stencil{m, extensions.NewHost(), mods, true, make(map[string]interface{})}
 }
 
 // Stencil provides the basic functions for
@@ -35,6 +35,13 @@ type Stencil struct {
 
 	// modules is a list of modules used in this stencil render
 	modules []*modules.Module
+
+	// isFirstPass denotes if the renderer is currently in first
+	// pass mode
+	isFirstPass bool
+
+	// sharedData stores data that is injected by templates from modules
+	sharedData map[string]interface{}
 }
 
 // RegisterExtensions registers all extensions on the currently loaded
@@ -87,6 +94,17 @@ func (s *Stencil) Render(ctx context.Context, log logrus.FieldLogger) ([]*Templa
 		return nil, err
 	}
 
+	// IDEA(jaredallard): Consider sharing this somehow, it's going
+	// to be pretty memory inefficient
+	//
+	// copy the templates into a separate slice for the first pass
+	// to prevent extra mutations
+	firstPass := make([]*Template, len(tplfiles))
+	for i, t := range tplfiles {
+		nt := *t
+		firstPass[i] = &nt
+	}
+
 	vals := NewValues(ctx, s.m)
 	tpls := make([]*Template, 0)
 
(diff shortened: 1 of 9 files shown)


pkg/configuration/configuration.go
```diff
@@ -118,6 +118,24 @@ type TemplateRepositoryManifest struct {
 
 	// Arguments are a declaration of arguments to the template generator
 	Arguments map[string]Argument
+
+	// ModuleBlocks contains all of the module to module blocks that
+	// this module provides to other modules
+	ModuleBlocks struct {
```
jaredallard, line 124: Not sure if I want to go with static defining of module blocks, so may remove this. It's a future optimization, IMO.

internal/codegen/stencil.go
```diff
@@ -23,7 +23,7 @@ import (
 
 // NewStencil creates a new, fully initialized Stencil renderer function
 func NewStencil(m *configuration.ServiceManifest, mods []*modules.Module) *Stencil {
-	return &Stencil{m, extensions.NewHost(), mods}
+	return &Stencil{m, extensions.NewHost(), mods, true, make(map[string]interface{})}
```
george-e-shaw-iv, line 26: we should initialize with named struct fields - it's more readable especially with this many fields and results in no performance hits (compiler optimizes for you)

george-e-shaw-iv (commented):
Honestly going to have to see an example of how this works in order to grok what is actually going on here.

internal/codegen/stencil.go
jaredallard: Makes sense!

internal/codegen/stencil_test.go
```diff
@@ -33,4 +37,82 @@ func TestBasicE2ERender(t *testing.T) {
...
+	fs := memfs.New()
+	ctx := context.Background()
+
+	// create a stub template
+	f, _ := fs.Create("manifest.yaml")
+	f.Write([]byte("name: testing\npostRunCommand:\n- command: echo \"hello\""))
+	f.Close()
+
+	nullLog := logrus.New()
+	logrus.SetOutput(io.Discard)
```
jaredallard, line 103:
```suggestion
	nullLog.SetOutput(io.Discard)
```

internal/codegen/tpl_stencil.go
```diff
@@ -20,21 +26,82 @@ type TplStencil struct {
...
+		s.s.sharedData[k] = data
+		return "", nil
+	}
+
+	return "", mergo.Merge(s.s.sharedData[k], data, mergo.WithAppendSlice)
+}
+
 // Arg returns the value of an argument in the service's
 // manifest.
 // Note: Only the top-level arguments are supported.
```
jaredallard, line 64:
```suggestion
```

internal/dotnotation/dotnotation.go
```diff
@@ -0,0 +1,45 @@
...
+// Package dotnotation implements a dotnotation (hello.world) for
+// accessing fields within a map[string]interface{}
+package dotnotation
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+)
+
```
jaredallard, line 15:
```suggestion
// Get looks up an entry in data by parsing the "key" into deeply nested keys, traversing it by "dots" in the key name.
```

internal/codegen/stencil.go
```diff
@@ -99,9 +117,17 @@ func (s *Stencil) Render(ctx context.Context, log logrus.FieldLogger) ([]*Templa
 		}
 	}
 
-	// Now we render each file
+	// Render the first pass, this is used to populate shared data
```
vk-outreach, line 120: Instead of two fixed passes and forcibly re-rendering every template, I suggest a slightly different technique:

With each module hook, track the templates that read from it (hook subscribers).
When  a hook is modified (i.e. AddModuleHook is called but with different contents than before), loop through all subscribers and mark them `dirty`.

When a pass of rendering is done, repeat it only for dirty templates.     We can support only two passes and if there are dirty templates still abort with error that templates are abusing hooks incorrectly.

jaredallard: I figured you might say this, I'll get it on it :D

vk-outreach: If you want a quick & dirty version, you can do two hardcoded passes, with the second pass only going through dirty templates.  And you can simplify dirty template tracking by only marking a template dirty if it called any code to fetch any hook (no need to track full subscribers list)

jaredallard: Going to implement this in a future PR -- I need to get some templates in (Monday at the latest, just need to get this stencil version out and the other branch)

PR #67, opened 2022-03-29
fix: use the correct 'its' as necessary

malept:

Extended version of #66


jaredallard (approved):
mark showing off his grammar skills ❤️

PR #71, opened 2022-04-04
feat(stencil): prompt on major version upgrades

jaredallard:

This PR adds major version upgrade prompts to stencil for modules. This displays the release notes and confirms with the user, if the say no they will need to run stencil with `--frozen-lockfile`

## Jira ID

[XX-XX]


cmd/stencil/create_templaterepository.go
```diff
@@ -11,7 +11,7 @@ import (
 
 	"github.com/getoutreach/stencil/pkg/configuration"
 	"github.com/urfave/cli/v2"
-	"gopkg.in/yaml.v3"
+	"gopkg.in/yaml.v2"
```
george-e-shaw-iv, line 14: wat

george-e-shaw-iv (approved):

cmd/stencil/create_templaterepository.go
jaredallard: the v3 one still isn't done

PR #73, opened 2022-04-05
fix(stencil): use templates/ as the base directory if present

jaredallard:

Keeping templates in the root has been... a big mistake, so, instead, we'll place them in `templates/` and keep the behavior but treating that as a "chroot".

## Jira ID

[DTSS-0]

### DIFF
--- a/internal/codegen/stencil.go
+++ b/internal/codegen/stencil.go
@@ -180,13 +180,17 @@ func (s *Stencil) getTemplates(ctx context.Context, log logrus.FieldLogger) ([]*
 
 		// default to templates/, but if it's not present fallback to
 		// the root w/ a warning
-		baseDir := "templates"
-		if inf, err := fs.Stat(baseDir); err != nil || !inf.IsDir() {
-			baseDir = ""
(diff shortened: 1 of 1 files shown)


internal/codegen/stencil.go
```diff
@@ -177,7 +177,14 @@ func (s *Stencil) getTemplates(ctx context.Context, log logrus.FieldLogger) ([]*
 		}
 
 		log.Debugf("Discovering templates from module %q", m.Name)
-		err = util.Walk(fs, "", func(path string, inf os.FileInfo, err error) error {
+		baseDir := ""
```
ChrisPRobinson, line 180: Maybe good to offer a warning of this.

george-e-shaw-iv, line 180:
```suggestion
		var baseDir string
```

jaredallard: don't tell me how to LIVE MY LIFE

george-e-shaw-iv: https://go-proverbs.github.io/

5th one down

george-e-shaw-iv (approved):

PR #74, opened 2022-04-08
fix(git): query remote for HEAD branch

jaredallard:

I added `-n` because it doesn't require network connectivity and seemed to still work, but apparently it breaks sometimes when a remote has never been queried before, so, we remove it again.

## Jira ID

[XX-XX]

### DIFF
--- a/internal/codegen/values_test.go
+++ b/internal/codegen/values_test.go
@@ -42,8 +42,8 @@ func TestValues(t *testing.T) {
 	assert.NilError(t, err, "expected worktree.Commit() not to fail")
 
 	err = wrk.Checkout(&gogit.CheckoutOptions{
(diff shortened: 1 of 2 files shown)


internal/codegen/values_test.go
```diff
@@ -56,7 +56,7 @@ func TestValues(t *testing.T) {
 	vals := NewValues(context.Background(), sm)
 	assert.DeepEqual(t, &Values{
 		Git: &git{
-			Ref:           plumbing.NewBranchReferenceName("main").String(),
```
george-e-shaw-iv, line 59: should u maybe just rewrite this to test both with a table test or something or does it matter

george-e-shaw-iv (approved):

PR #77, opened 2022-04-20
feat: fromYAML, file.Static, store args on template

jaredallard:

Title says, this PR does :tada:.

`file.Static` is a wrapper around `file.Skip` that calls it if the file already exists. `fromYaml` is for parsing YAML files (sub-templates, hooks w/e) and processing it.

Storing template args is so that `ApplyTemplate` can default to passing the already set args to sub-templates

## Jira ID

[XX-XX]


jaredallard:
This replaces https://github.com/getoutreach/stencil/pull/76

internal/codegen/functions_test.go
```diff
@@ -31,3 +31,26 @@ func Example_quotejoinstrings() {
 	// Output:
 	// "a" "b" "c"
 }
+
+func Example_toYaml() {
```
ChrisPRobinson, line 35: "// This is not a general-purpose YAML parser, and will not parse all valid
// YAML documents."

Which type of documents does it not parse? Would it be helpful to add an example where this will fail?  Perhaps this might help people, though unsure, just brainstorming

ChrisPRobinson, line 35: "// This is not a general-purpose YAML parser, and will not parse all valid
// YAML documents."

Which type of documents does it not parse? Would it be helpful to add an example where this will fail?  Perhaps this might help people, though unsure, just brainstorming

internal/codegen/tpl_stencil.go
```diff
@@ -138,6 +138,9 @@ func (s *TplStencil) ApplyTemplate(name string, dataSli ...interface{}) (string,
 	var data interface{}
 	if len(dataSli) == 1 {
 		data = dataSli[0]
+	} else {
+		// If no data was passed, pass through the values of the parent template
+		data = s.t.args
```
ChrisPRobinson, line 143: Is there a demonstration that uses this functionality in a test?

ChrisPRobinson, line 143: Is there a demonstration that uses this functionality in a test?

internal/codegen/values.go
```diff
@@ -60,6 +60,9 @@ type git struct {
 type config struct {
 	// Name is the name of this repository
 	Name string
+
+	// Versions are versions of applications to be used over baked in defaults
+	Versions map[string]string
```
ChrisPRobinson, line 65: Also curious of usage here as well.

ChrisPRobinson, line 65: Also curious of usage here as well.

ChrisPRobinson (approved):
LGTM though hard to understand how some of these bits are used and in what scenarios.

internal/codegen/functions_test.go
jaredallard: Yeah fair, I copied this from helm... I'm not sure what it couldn't parse honestly

internal/codegen/tpl_stencil.go
jaredallard: There is not, I'll add one!

internal/codegen/values.go
jaredallard: This is kinda a placeholder for stencil-golang, will think about it more as I work on it

PR #78, opened 2022-04-22
fix: only support []interface{} for hooks, module hooks render properly

jaredallard:

This PR changes module hooks to only accept slices and to return `[]interface{}` instead of `interface{}`. This brings us better in line with the true value of hooks, which is for multiple items.

Module hooks previously erroneously shared a slice between first pass and second pass. This would result in module hooks working but not persisting their contents because of how the template default file system works. This is now fixed by clearing `Files` between runs, which is the only mutated field of a template struct. This is also more memory efficient. This will be improved upon when we move to a selective rerender system in the future.

Debugging has also been beefed up by adding module hook logs into the `--debug` logs.

## Jira ID

[XX-XX]

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,7 +12,8 @@
       "envFile": "${workspaceRoot}/.vscode/private.env",
       "go.testEnvFile": "${workspaceRoot}/.vscode/private.env",
       "program": "${workspaceRoot}/cmd/stencil/",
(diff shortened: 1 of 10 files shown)


george-e-shaw-iv (approved):

jaredallard:
I'll merge this once stencil-base/circle has been updated since this is technically a breaking change.

PR #82, opened 2022-04-29
feat: prerelease templaterepositories by default, support module version override

jaredallard:

This PR updates `create templaterepository` to default to pre-releasing them. I also fixed an issue with top level modules being unable to override dependency versions, with a note on a better path forward for the future.

## Jira ID

[XX-XX]

### DIFF
--- a/internal/modules/modules.go
+++ b/internal/modules/modules.go
@@ -77,8 +77,9 @@ func getModulesForService(ctx context.Context, sm *configuration.ServiceManifest
 				continue
 			}
 
-			// GC the old module
-			modules[d.Name] = nil
+			// GC the old module, we're not going to use it, and it'll be
+			// replaced later.
+			delete(modules, d.Name)
 		}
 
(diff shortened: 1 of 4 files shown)


internal/modules/modules.go
```diff
@@ -54,9 +55,30 @@ func getModulesForService(ctx context.Context, sm *configuration.ServiceManifest
...
+			}
+
+			// The new version is less than the one we already found, so
+			// we skip downloading the module here.
+			if newV.LTE(curV) {
+				continue
+			}
+
+			// GC the old module
+			modules[d.Name] = nil
```
george-e-shaw-iv, line 81: delete(modules, d.Name) would be better

george-e-shaw-iv (approved):

internal/modules/modules.go
jaredallard: good call

PR #86, opened 2022-05-03
feat(extensions): serialize data via JSON

jaredallard:

This PR reworks the extension framework to properly support all types via serialization to JSON. Gob requires registration of types on both the client and server-side, which we can't currently support. Using JSON also allows us to drop the type requirements which simplifies the implementation of plugins and the plugin executor.

## Jira ID

[XX-XX]


pkg/extensions/apiv1/apiv1.go
```diff
@@ -2,9 +2,18 @@
...
-// Package apiv1 implements the extension API for stencil
+// Package apiv1 implements the bridge between a extension and go-plugin
+// providing most of the implementation for the extension if it's
+// written in Go.
 package apiv1
 
+import "encoding/gob"
+
+// init registers known types
+func init() { //nolint:gochecknoinits
```
george-e-shaw-iv, line 13: add a // Why so lintroller doesn't scream

george-e-shaw-iv (approved):

PR #101, opened 2022-05-31
feat: support prerelease on module (DTSS-1924)

jaredallard:
**What this PR does**: This PR adds a `--use-prerelease` flag to use prereleases across all modules, or the ability to set `prerelease` to `true` on the module in the `service.yaml`


internal/modules/module.go
```diff
@@ -52,37 +53,51 @@ type Module struct {
...
-		return "", fmt.Errorf("invalid module path %q, expected github.com/<org>/<repo>", name)
+		return "", fmt.Errorf("invalid module path %q, expected github.com/<org>/<repo>", tr.Name)
 	}
 
 	gh, err := github.NewClient(github.WithAllowUnauthenticated())
 	if err != nil {
 		return "", err
 	}
 
+	if tr.Prerelease { // Use the newest, first, release.
```
ChrisPRobinson, line 69: More of a question on the github api. GetLatestRelease will not return a latest RC?

internal/modules/module.go
```diff
@@ -52,37 +53,51 @@ type Module struct {
...
+		rels, _, err := gh.Repositories.ListReleases(ctx, paths[1], paths[2], &gogithub.ListOptions{
+			PerPage: 1,
+		})
+		if err != nil {
+			return "", errors.Wrap(err, "failed to get releases")
+		}
+		if len(rels) != 1 {
+			return "", errors.New("failed to get latest release (returned >1 release)")
```
ChrisPRobinson, line 77: Again it might make sense to add the module name here.

internal/modules/module.go
jaredallard: `GetLatestRelease` returns the latest _released_ version. https://docs.github.com/en/rest/releases/releases#get-the-latest-release

internal/modules/module.go
```diff
@@ -52,37 +53,51 @@ type Module struct {
...
 	if err != nil {
 		return "", err
 	}
 
+	if tr.Prerelease { // Use the newest, first, release.
+		rels, _, err := gh.Repositories.ListReleases(ctx, paths[1], paths[2], &gogithub.ListOptions{
+			PerPage: 1,
+		})
+		if err != nil {
+			return "", errors.Wrap(err, "failed to get releases")
```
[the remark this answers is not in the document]
jaredallard:
```suggestion
			return "", errors.Wrapf(err, "failed to get releases for module %q", tr.Name)
```

PR #107, opened 2022-06-07
fix: add --allow-major-version-upgrades to skip confirmation, fail when stdin unavailable

jaredallard:
**What this PR does**: When we're not a terminal we can't approve changes, so we fail. This adds a `--allow-major-version-upgrades` flag to skip confirmation checking.


george-e-shaw-iv (approved):
Should be flag but take the green check mark

PR #117, opened 2022-07-06
fix(DTSS-1817): Address dangling Block/EndBlock statements with specific errors

deregtd:
## Jira ID

[XX-XX]

### DIFF
--- a/internal/codegen/blocks.go
+++ b/internal/codegen/blocks.go
@@ -46,7 +46,7 @@ func parseBlocks(filePath string) (map[string]string, error) {
 			case "Block":
 				blockName := matches[3]
 				if curBlockName != "" {
-					return nil, fmt.Errorf("invalid Block when already inside of a block, at %s:%d", filePath, i)
+					return nil, fmt.Errorf("invalid Block when already inside of a block, at %s:%d", filePath, i+1)
 				}
 				curBlockName = blockName
 			case "EndBlock":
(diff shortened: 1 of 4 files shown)


internal/codegen/blocks.go
```diff
@@ -46,20 +46,20 @@ func parseBlocks(filePath string) (map[string]string, error) {
 			case "Block":
 				blockName := matches[3]
 				if curBlockName != "" {
-					return nil, fmt.Errorf("invalid Block when already inside of a block, at %s:%d", filePath, i)
+					return nil, fmt.Errorf("invalid Block when already inside of a block, at %s:%d", filePath, i+1)
```
artemchernyak-outreach, line 49: Changes look good, but would be nice to add a test case for this scenario too.

internal/codegen/blocks_test.go
```diff
@@ -15,3 +15,15 @@ func TestParseBlocks(t *testing.T) {
...
+
+func TestDanglingBlock(t *testing.T) {
+	_, err := parseBlocks("testdata/danglingblock-test.txt")
+	assert.Error(t, err, "found dangling Block (dangles) in testdata/danglingblock-test.txt", "expected parseBlocks() to fail")
+}
+
+func TestDanglingEndBlock(t *testing.T) {
+	_, err := parseBlocks("testdata/danglingendblock-test.txt")
+	assert.Error(t, err,
+		"invalid EndBlock, found EndBlock with name \"dangles\" while inside of block with name \"\", at testdata/danglingendblock-test.txt:8",
```
artemchernyak-outreach, line 27: Looks like this is getting the wrong error. It should get `invalid EndBlock when not inside of a block ...`.

artemchernyak-outreach (changes requested):

internal/codegen/blocks_test.go
deregtd: done

internal/codegen/blocks.go
deregtd: done

artemchernyak-outreach (approved):
Looks great!

PR #129, opened 2022-08-18
feat(arguments): support from to alias to another module's argument

jaredallard:

This PR adds a new `from` field to the argument module spec. `from` alias the current argument name, based on the key in the `arguments` map, to another module's argument spec.

**Example**:

```yaml
name: test-0
arguments:
  hello:
    from: test-1

name: test-1
arguments:
  hello:
    schema:
      type: string
```

`test-0.hello` would inherti all the properties of the `test-1.hello` argument.

There's a few special rules here:

 * When `from` is set, no other keys apply. e.g. if you have `from` and `schema`, `schema` would be ignored in favor of the `schema` in the module referenced. There is no sort of merging.
 * An argument using `from` must declare it as a dependency in the `modules` list for `from` to work, otherwise a runtime-error will occur

## Jira ID

[XX-XX]

### DIFF
--- a/internal/codegen/tpl_stencil_arg.go
+++ b/internal/codegen/tpl_stencil_arg.go
@@ -1,0 +1,222 @@
+// Copyright 2022 Outreach Corporation. All Rights Reserved.
+
+// Description: This file contains the code for the stencil.Arg
+// template function.
+
+package codegen
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/getoutreach/stencil/internal/dotnotation"
+	"github.com/getoutreach/stencil/pkg/configuration"
+	"github.com/pkg/errors"
+	"github.com/santhosh-tekuri/jsonschema/v5"
+)
+
+// Arg returns the value of an argument in the service's manifest
+//
+//	{{- stencil.Arg "name" }}
+func (s *TplStencil) Arg(pth string) (interface{}, error) {
+	if pth == "" {
+		// Deprecated: Arg(pth) will be required in the future. See note on
+		// Args().
+		return s.Args(), nil
+	}
+
+	// This is a TODO because I don't know if template functions
+	// can even get a context passed to them
+	ctx := context.TODO()
+
+	mf, err := s.t.Module.Manifest(ctx)
+	if err != nil {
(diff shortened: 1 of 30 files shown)


jaredallard:
Doing this highlighted the need for better test helpers, this is the bulk of the work here. There's a lot more I'd like to do, but this suffices for now I think 😃

internal/codegen/tpl_stencil_arg.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/pkg/errors"
+	"github.com/santhosh-tekuri/jsonschema/v5"
+)
+
+// Arg returns the value of an argument in the service's manifest
+//
+//	{{- stencil.Arg "name" }}
+func (s *TplStencil) Arg(pth string) (interface{}, error) {
+	if pth == "" {
+		// Deprecated: Arg(pth) will be required in the future. See note on
```
george-e-shaw-iv, line 25: Should probably track depreciations somewhere other than in a comment - they're bound to be forgotten about

internal/codegen/tpl_stencil_arg.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	if !foundModuleInDeps {
+		return nil, fmt.Errorf(
+			"module %q argument %q references an argument in module %q, but doesn't list it as a dependency",
+			s.t.Module.Name, pth, arg.From,
+		)
+	}
+
+	// Get the manifest for the referenced module
+	var fromMf *configuration.TemplateRepositoryManifest
```
george-e-shaw-iv, line 151: This might be a preemptive comment (you might've already done this) but I'd call out in the docs that from isn't transitive - it can only go one layer deep. I don't know why we'd ever end up there tho

george-e-shaw-iv (approved):
Not important comments

internal/codegen/tpl_stencil_arg.go
jaredallard: Agreed, I'll move it to the docs

internal/codegen/tpl_stencil_arg.go
jaredallard: Good call

PR #133, opened 2022-09-02
fix(extensions): close extensions after run (DT-2616)

jaredallard:

This PR fixes an issue where stencil wouldn't terminate native extensions after it had finished running. I also updated all the dependencies, and handled the breaking change in the updater component.

## Jira ID

[DT-2616]

## Notes for your reviewers

[DT-2616]: https://outreach-io.atlassian.net/browse/DT-2616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/extensions/extensions.go
```diff
@@ -156,53 +162,73 @@ func (h *Host) getExtensionPath(version, name, repo string) string {
...
+	dlPath := h.getExtensionPath(version, name)
 	if info, err := os.Stat(dlPath); err == nil && info.Mode() == 0o755 {
 		return dlPath, nil
 	}
 
-	// Binary for plugin at version we want doesn't exist on disk, need to download.
-	bin, cleanup, err := gh.DownloadRelease(ctx, rel, repo, repo)
+	h.log.WithField("version", version).WithField("repo", repoURL).Debug("Downloading native extension")
+	a, archiveName, _, err := release.Fetch(ctx, token, &release.FetchOptions{
+		AssetName: filepath.Base(name) + "_" + strings.TrimPrefix(version, "v") + "_" + runtime.GOOS + "_" + runtime.GOARCH + ".tar.gz",
```
jkinkead, line 192: This would probably be more readable as a `fmt.Sprintf`.

Although - is there not a shared function to build something like this yet? It seems like there should be one somewhere already.

pkg/extensions/extensions.go
```diff
@@ -156,53 +162,73 @@ func (h *Host) getExtensionPath(version, name, repo string) string {
...
+	f, err := os.Create(dlPath)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to create file")
+	}
+	defer f.Close()
+
+	if _, err := io.Copy(f, bin); err != nil {
+		return "", errors.Wrap(err, "failed to download binary")
+	}
+	f.Close()
```
jkinkead, line 214: Already deferred on line 209?

pkg/extensions/extensions.go
```diff
@@ -156,53 +162,73 @@ func (h *Host) getExtensionPath(version, name, repo string) string {
...
-		return "", errors.Wrap(err, "failed to download extension")
+		return "", errors.Wrap(err, "failed to fetch release")
 	}
-	defer cleanup()
 
-	// Move the downloaded release from where the updater put it to where we need it
-	// for stencil.
-	if err := os.Rename(bin, dlPath); err != nil {
-		return "", errors.Wrap(err, "failed to move downloaded extension")
+	bin, _, err := archive.Extract(ctx, archiveName, a, archive.WithFilePath(filepath.Base(name)))
```
jkinkead, line 200: Is this buffering the whole binary into memory before writing to disk?

If so, can you stream to disk instead?

jkinkead (approved):
LGTM with a couple small suggestions.

pkg/extensions/extensions.go
jaredallard: Yeah, this is something I've recognized as something I need to build out on the new updater rewrite I did. It's been in the back of my head for a bit, let me just go ahead and create a ticket so I don't forget and we do the work soon

pkg/extensions/extensions.go
jaredallard: Yep, though technically open file handles can break things like `Chmod` so I'm doing this out of an abundance of caution. I hate duplicating it when I'm already using `defer` but really the `defer` call there is the "fallback" in case we hit one of the returns.

pkg/extensions/extensions.go
jaredallard: Nope, this is all streaming! Github -> archive -> binary -> disk, it's all being done on the fly here 🎉

jaredallard: (it was one of the things I did when re-writing the updater last week that I'm super happy with)

pkg/extensions/extensions.go
jaredallard: Tracking that in [DT-2786]

jaredallard:
Depends on this PR: https://github.com/getoutreach/devbase/pull/296

PR #134, opened 2022-09-02
feat(tpl): add stencil.ReadBlocks to parse blocks from arbitrary files

jaredallard:

This PR adds `stencil.ReadBlocks "path"` to read the blocks out of an arbitrary path. This is useful for moving files from one place to the other, or some other attempt at reading the blocks of another file.

## Jira ID

[XX-XX]

### DIFF
--- a/internal/codegen/tpl_stencil_test.go
+++ b/internal/codegen/tpl_stencil_test.go
@@ -1,0 +1,64 @@
+// Copyright 2022 Outreach Corporation. All Rights Reserved.
+
+// Description: This file contains the public API for templates
+// for stencil
+
+package codegen
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestTplStencil_ReadBlocks(t *testing.T) {
+	type fields struct {
+	}
+	type args struct {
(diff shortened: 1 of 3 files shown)


internal/codegen/tpl_stencil_test.go
```diff
@@ -0,0 +1,64 @@
...
+				"helloWorld": "Hello, world!",
+				"e2e":        "content",
+			},
+		},
+		{
+			name: "should error on out of chroot path",
+			args: args{
+				fpath: "../testdata/blocks-test.txt",
+			},
+			wantErr: true,
```
jkinkead, line 41: Given that this is a security check, it would be better if you verified the error type / contents and not just that one was returned.

jaredallard: Hm yeah I can do that. These test were 99% generated by VSCode so it doesn't support that OOTB but I can quickly edit the table tests to support checking the error type / message (if the type isn't available)

george-e-shaw-iv (approved):

PR #137, opened 2022-09-09
feat(modules): version resolver w/ constraint support (DT-2120)

jaredallard:

This PR adds a version constraint aware resolver to stencil. This essentially turns the existing `version` selector into a constraint section that can be a specific version, or a semantic version constraint.

Currently this does _not_ support branches. The underlying version resolver will be updated to support that before this PR is merged, which should automatically bring support for branches here.

## Jira ID

[DT-2120]

## Notes for your reviewers

Tests will be failing until I un-replace gobox.

[DT-2120]: https://outreach-io.atlassian.net/browse/DT-2120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/modules/modules.go
```diff
@@ -10,123 +10,174 @@ package modules
...
+func getLatestModuleForConstraints(ctx context.Context, token cfg.SecretData,
+	m *resolveModule, resolved map[string]*resolvedModule) (string, error) {
+	constraints := resolved[m.conf.Name].constraints
+	if len(constraints) == 0 {
+		constraints = make([]constraint, 0)
+	}
+
+	// if we have a constraint, use it to resolve the version
+	if m.conf.Version != "" {
+		constraints = append(constraints, constraint{
```
jaredallard, line 156: Thinking maybe even if we don't have a constraint we should still "commit it" as a `*` 🤔

internal/modules/modules.go
```diff
@@ -10,123 +10,174 @@ package modules
...
+		resolved[m.conf.Name].constraints = constraints
+	}
+
+	constraintsStr := make([]string, len(constraints))
+	for i := range constraints {
+		constraintsStr[i] = constraints[i].str
+	}
+
+	v, err := resolver.Resolve(ctx, token, &resolver.Criteria{
+		URL:         "https://" + m.conf.Name,
```
jaredallard, line 169: Thinking I need to plumb `uri` here for replacements that are URLs to work

internal/modules/modules.go
jaredallard: More thinking, instead of specifically tracking just constraints, I think we should log the entire tree of the resolution. That'll make some errors more clear, and get this as well (it's just type renaming I guess)

internal/modules/modules.go
```diff
@@ -10,123 +10,174 @@ package modules
...
+	}
+
+	v, err := resolver.Resolve(ctx, token, &resolver.Criteria{
+		URL:         uri,
+		Channel:     m.conf.Channel,
+		Constraints: constraintsStr,
+	})
+	if err != nil {
+		errorString := ""
+		for i := range constraints {
```
jaredallard, line 175: I'll want to expose the resolution information in a different format than the error string, but that'll increase the scope of this work a decent amount, so for now this is why this is going in the error string.

If future me gets mad about this, I'm sorry 😢

pkg/configuration/configuration.go
```diff
@@ -74,16 +73,23 @@ type TemplateRepository struct {
...
+	// Channel is the channel to use for updates to this module
+	// Defaults to "stable"
+	Channel string `yaml:"channel,omitempty"`
+
 	// Version is a semantic version or branch of the template repository
 	// that should be downloaded if not set then the latest version is used.
-	// Note: Setting this equates to pinning the versions, this is not recommended.
+	//
+	// Version can also be a constraint as supported by the underlying resolver:
+	// !!! <TODO link to gobox/pkg/cli/updater/resolver>
```
jaredallard, line 92: This TODO is known, this PR has to land first: https://github.com/getoutreach/gobox/pull/219

artemchernyak-outreach (approved):

PR #139, opened 2022-09-13
fix(modules): properly persist channel resolve information

jaredallard:

The version resolver does a great job of persisting constraints across version resolves, but it did not calculate channels properly. This fixes that to calculate channels, and error when more than one channel is requested (since we do not know the logical flow of a channel in terms of promotion)

## Jira ID

[DT-2808]

## Notes for your reviewers

[DT-2808]: https://outreach-io.atlassian.net/browse/DT-2808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/stencil/stencil.go
+++ b/cmd/stencil/stencil.go
@@ -110,6 +110,7 @@ func main() {
 		},
 		&cli.BoolFlag{
 			Name:    "debug",
+			Usage:   "Enables debug logging for version resolution, template render, and other useful information",
 			Aliases: []string{"d"},
 		},
 		///EndBlock(flags)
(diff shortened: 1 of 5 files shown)


cmd/stencil/stencil.go
```diff
@@ -108,6 +108,10 @@ func main() {
 			Name:  "allow-major-version-upgrades",
 			Usage: "Allow major version upgrades without confirmation",
 		},
+		&cli.BoolFlag{
+			Name:    "debug",
+			Aliases: []string{"d"},
```
george-e-shaw-iv, line 113: @jkinkead wants u to document ur flags

PR #143, opened 2022-09-14
feat(blocks): use safer block format (v2)

jaredallard:

This PR adds a new block format (v2) to get around issues with comment fomatters that would break the traditional `///`, among other cases, blocks.

The new block format looks like so:

```go
// <<Stencil::Block(name)>>
fmt.Println("Block stuff does block things")
// <</Stencil::Block>>
```

## Jira ID

[DT-0]

### DIFF
--- a/internal/codegen/blocks.go
+++ b/internal/codegen/blocks.go
@@ -9,9 +9,13 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strings"
 
 	"github.com/pkg/errors"
 )
+
+// endStatement is a constant for the end of a statement
+const endStatement = "EndBlock"
 
 // blockPattern is the regex used for parsing block commands.
 // For unit testing of this regex and explanation, see https://regex101.com/r/EHkH5O/1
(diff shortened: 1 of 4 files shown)


internal/codegen/blocks.go
```diff
@@ -14,12 +14,16 @@ import (
 )
 
 // blockPattern is the regex used for parsing block commands.
-// For unit testing of this regex and explanation, see https://regex101.com/r/nFgOz0/1
+// For unit testing of this regex and explanation, see https://regex101.com/r/EHkH5O/1
```
jaredallard, line 17: Need to reset this line

internal/codegen/blocks.go
```diff
@@ -33,6 +41,45 @@ func parseBlocks(filePath string) (map[string]string, error) {
...
+						return nil, fmt.Errorf("line %d: expected no arguments to <</Stencil::Block>>", i+1)
+					}
+
+					v2Matches[4] = fmt.Sprintf("(%s)", curBlockName)
+				} else if cmd == endStatement {
+					// If it's not a closing tag, but the command is EndBlock,
+					// we should error. This is because we don't want to
+					// allow users to use the old EndBlock command
+					// without a closing tag
+					return nil, errors.Errorf("line %d: EndBlock command must be used with a closing tag", i+1)
```
jaredallard, line 71:
```suggestion
					return nil, errors.Errorf("line %d: <<Stencil::EndBlock>> should be <</Stencil::Block>>", i+1)
```

I think this is a more clear error message

internal/codegen/blocks.go
```diff
@@ -33,6 +41,45 @@ func parseBlocks(filePath string) (map[string]string, error) {
...
+				}
+
+				// fake the old matches format so we can reuse the same code
+				matches = []string{
+					v2Matches[0],
+					v2Matches[1],
```
jkinkead, line 77: Are you using this capture at all? Seems like you could uncapture it if you were feeling tidy.

jkinkead (approved):
LGTM!

internal/codegen/blocks.go
jaredallard: Which one? It's not super clear from the comment here which it's referring too 😢

jkinkead: The first capture group being pulled out of `v2matches`. You're capturing the comment leader but never referencing it, AFAICT?

PR #148, opened 2022-09-20
feat(codegen/stencil): add stencil.Exists tpl function

jaredallard:

This PR adds a `stencil.Exists` template function to check if a file exists in the current directory. The logic behind this function is that when a [go-template function returns an error, the template execution is halted](https://pkg.go.dev/text/template#FuncMap). This makes it painful to use `stencil.ReadFile` when doing custom "static-like" logic. While it'd be more effective to add a `catch` or `stencil.Call` function to handle that error for the user, we have more we'll want to do there in the future for a `ApplyTemplate` revamp to better support template helpers, so I'm not doing that yet.

## Jira ID

[DT-2808]

## Notes for your reviewers

[DT-2808]: https://outreach-io.atlassian.net/browse/DT-2808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/codegen/tpl_stencil.go
+++ b/internal/codegen/tpl_stencil.go
@@ -14,6 +14,7 @@ import (
 	"reflect"
 
 	"github.com/davecgh/go-spew/spew"
+	"github.com/go-git/go-billy/v5"
 	"github.com/go-git/go-billy/v5/osfs"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


internal/codegen/tpl_stencil.go
```diff
@@ -119,23 +120,43 @@ func (s *TplStencil) Args() map[string]interface{} {
...
+	return string(b), nil
+}
+
+// Exists returns true if the file exists in the current directory
+//
+//	{{- if stencil.Exists "myfile.txt" }}
+//	{{ stencil.ReadFile "myfile.txt" }}
+//	{{- end }}
+func (s *TplStencil) Exists(name string) bool {
+	f, ok := s.exists(name)
```
jaredallard, line 142: I didn't see much value in adding a test for this because it's using a function already covered (Readfile -> exists). I can add one if someone wants though

internal/codegen/tpl_stencil.go
```diff
@@ -119,23 +120,43 @@ func (s *TplStencil) Args() map[string]interface{} {
...
+	f.Close() // close the file handle, since we don't need it
```
chatchai-outreach, line 143: f.Close() if ok?

jaredallard: yeah, i fixed this locally and then got pulled into a meeting. Thank you!

chatchai-outreach (approved):

artemchernyak-outreach (approved):
LGTM

PR #150, opened 2022-09-20
fix(gobox): update gobox for latest updater features

jaredallard:

Does what it says on the tin.

## Jira ID

[XX-XX]


artemchernyak-outreach (approved):

jaredallard:
Ugh, there is some sort of regression with this change. Will need to debug that

PR #158, opened 2022-09-23
fix: small UX/debugging improvements

### LINKED ISSUE #157: stencil describe can't find file
Looks like it is looking for the exact path from stencil.lock, but the behavior is a little surprising, as I would expect it to accept a reasonable looking file path as well

```
❯ stencil describe cmd/smartstorecdctestconsumer/smartstore.go
cmd/smartstorecdctestconsumer/smartstore.go was created by module https://github.com/getoutreach/stencil-smartstore (template: cmd/smartstore.go.tpl)

❯ stencil describe ./cmd/smartstorecdctestconsumer/smartstore.go
ERRO[0000] failed to run: file "./cmd/smartstorecdctestconsumer/smartstore.go" isn't created by stencil
```

jaredallard:

This PR adds `SkipReason` (which is already in the function signature, but not used) in debug logs, and static log debugging. I also fixed #157.

## Jira ID

[DT-2902]

## Notes for your reviewers

[DT-2902]: https://outreach-io.atlassian.net/browse/DT-2902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/stencil/describe.go
+++ b/cmd/stencil/describe.go
@@ -22,35 +22,52 @@ func NewDescribeCmd() *cli.Command {
 		Name:        "describe",
 		Description: "Print information about a known file rendered by a template",
 		Action: func(c *cli.Context) error {
-			l, err := stencil.LoadLockfile("")
-			if err != nil {
-				return errors.Wrap(err, "failed to load lockfile")
+			if c.NArg() != 1 {
+				return errors.New("expected exactly one argument, path to file")
 			}
 
-			// make absolute so we can handle .. and other weird path things
-			// defaults to nothing if already absolute
-			filePath, err := filepath.Abs(c.Args().First())
-			if err != nil {
-				return errors.Wrap(err, "failed to get absolute path")
-			}
-
-			// convert absolute -> relative
-			cwd, err := os.Getwd()
-			if err != nil {
-				return errors.Wrap(err, "failed to get current working directory")
-			}
-			filePath = "." + strings.TrimPrefix(filePath, cwd)
-
-			// ensure that we don't have any weird path elements
-			relativeFilePath := filepath.Clean(filePath)
-			for _, f := range l.Files {
-				if f.Name == relativeFilePath {
-					fmt.Printf("%s was created by module https://%s (template: %s)\n", f.Name, f.Module, f.Template)
-					return nil
-				}
-			}
-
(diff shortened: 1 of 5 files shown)


cmd/stencil/describe.go
```diff
@@ -24,15 +27,30 @@ func NewDescribeCmd() *cli.Command {
...
+
+			// convert absolute -> relative
+			cwd, err := os.Getwd()
+			if err != nil {
+				return errors.Wrap(err, "failed to get current working directory")
+			}
+			filePath = "." + strings.TrimPrefix(filePath, cwd)
+
+			// ensure that we don't have any weird path elements
+			relativeFilePath := filepath.Clean(filePath)
```
jkinkead, line 45: Add a unit test for this new logic?

cmd/stencil/describe.go
```diff
@@ -24,15 +27,30 @@ func NewDescribeCmd() *cli.Command {
 				return errors.Wrap(err, "failed to load lockfile")
 			}
 
-			fileName := c.Args().First()
+			// make absolute so we can handle .. and other weird path things
+			// defaults to nothing if already absolute
+			filePath, err := filepath.Abs(c.Args().First())
```
jkinkead, line 32: Check to see if there is exactly one argument, and return an error if not?

cmd/stencil/describe.go
jaredallard: I knew you were going to ask that 😄

cmd/stencil/describe.go
```diff
@@ -19,20 +22,52 @@ func NewDescribeCmd() *cli.Command {
...
+}
+
+// describeFile prints information about a file rendered by a template
+func describeFile(filePath string) error {
+	l, err := stencil.LoadLockfile("")
+	if err != nil {
+		return errors.Wrap(err, "failed to load lockfile")
+	}
+
+	relativeFilePath, err := cleanPath(filePath)
```
jkinkead, line 60: Suggestion: stat the file before looking for it in the lockfile? It might be nice to have a separate "file does not exist" error.

jkinkead (approved):

PR #163, opened 2022-09-30
fix: frozen-lockfile properly handles transitive dependencies, tests

jaredallard:

Previously the logic for `--frozen-lockfile` would only iterate over the modules in the `modules` key of the `service.yaml`, which made sense because that's what's fed into the version resolver. This would then mutate that and set a constraint on the module being equal to the value found in the lockfile. 

This wouldn't work for transitive dependencies because the lockfile module didn't exist inside of the manifest. This fixes the logic to instead source modules from the lockfile, adding a top-level module if it's not found (which is a _bit_ of a hack, but is in line with how we do version feeding today).

I added tests and finalized the codecov integration with this repository as well.

## Jira ID

[DT-0]

### DIFF
--- a/internal/cmd/stencil/stencil.go
+++ b/internal/cmd/stencil/stencil.go
@@ -206,6 +206,12 @@ func (c *Command) useModulesFromLock() error {
 		// to =<version> so the resolver only considers
 		// the version from the lockfile.
 		for _, m := range c.manifest.Modules {
+			// Set channel and pre-release to false to avoid accidentally
+			// de-selecting the version from the lockfile.
+			m.Channel = ""
+			//nolint:staticcheck // Why: Resetting value to false
+			m.Prerelease = false
+
 			if m.Name == l.Name {
 				m.Version = l.Version
 			}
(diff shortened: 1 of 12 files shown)


artemchernyak-outreach (approved):

internal/cmd/stencil/stencil.go
```diff
@@ -152,31 +150,72 @@ func (c *Command) useModulesFromLock() error {
...
+
+	// iterate over all of the modules that are desired, if
+	// they are not in the lockfile, then the user is unable
+	// to use a frozen lockfile.
+	for _, m := range c.manifest.Modules {
+		if _, ok := lockfileModulesHM[m.Name]; !ok {
+			outOfSync = true
+			outOfSyncReasons = append(outOfSyncReasons,
+				fmt.Sprintf("module %s requested by service.yaml but is not in the lockfile", m.Name))
+			break
```
jkinkead, line 174: Why do you break here? Don't you want to gather all reasons?

jaredallard: Because Jesse, you're not the boss of me. o wait. No, in reality I changed this to gather multiple reasons and never removed the break. That's why :D

jkinkead: also seems like an opportunity for your unit tests to cover this case.

jaredallard: Github showed me this after merge :sob:

Since it's just logged right now I'd have to change the log messages, but yeah this is potentially something we could unit test. I feel _generally_ safe enough with it checking error cases right now though

PR #171, opened 2022-11-03
fix(jsonschema): provide more explicit schema error

samnguyen-outreach:

- If there is a single JSON schema validation error, a lot of the path information currently gets thrown away, so the user cannot determine where exactly in the service.yaml the validation is failing.  
- This removes the filtering in the case where there is just 1 error so that the entire detailed error is displayed

Before change (does not indicate any path information beyond the top-level service.yaml argument)
```
ERRO[0010] failed to run: run codegen: 
failed to render template "github.com/getoutreach/stencil-smartstore/_helpers.tpl": 
template: github.com/getoutreach/stencil-smartstore/_helpers.tpl:5:25:
executing "github.com/getoutreach/stencil-smartstore/_helpers.tpl" at <stencil.Arg>: 
error calling Arg: module "github.com/getoutreach/stencil-smartstore" argument "postgreSQL" validation failed: does not match pattern '^[a-z_\\-]+$'
```

After change
```
ERRO[0011] failed to run: run codegen: 
failed to render template "github.com/getoutreach/stencil-smartstore/_helpers.tpl": template: github.com/getoutreach/stencil-smartstore/_helpers.tpl:5:25: 
executing "github.com/getoutreach/stencil-smartstore/_helpers.tpl" at <stencil.Arg>: 
error calling Arg: module "github.com/getoutreach/stencil-smartstore" argument "postgreSQL" validation failed: 
jsonschema.Detailed{Valid:false, KeywordLocation:"", AbsoluteKeywordLocation:"file:///Users/sam.nguyen/src/stencil-smartstore/testapps/orgschemagrpc/manifest.yaml/arguments/postgreSQL#", InstanceLocation:"", Error:"", Errors:[]jsonschema.Detailed{jsonschema.Detailed{Valid:false, KeywordLocation:"/items/properties/name/pattern", AbsoluteKeywordLocation:"file:///Users/sam.nguyen/src/stencil-smartstore/testapps/orgschemagrpc/manifest.yaml/arguments/postgreSQL#/items/properties/name/pattern", InstanceLocation:"/0/name", Error:"does not match pattern '^[a-z_\\\\-]+$'", Errors:[]jsonschema.Detailed(nil)}}}
```

## Jira ID

[XX-XX]

## Notes for your reviewers

- I also tried JSON encoding the error but it gets escaped so there are a lot of `\` in the output

### DIFF
--- a/internal/codegen/tpl_stencil_arg.go
+++ b/internal/codegen/tpl_stencil_arg.go
@@ -205,17 +205,8 @@ func (s *TplStencil) validateArg(pth string, arg *configuration.Argument, v inte
 	if err := schema.Validate(v); err != nil {
 		var validationError *jsonschema.ValidationError
 		if errors.As(err, &validationError) {
-			// If there's only one error, return it directly, otherwise
-			// return the full list of errors.
-			errs := validationError.DetailedOutput().Errors
-			out := ""
-			if len(errs) == 1 {
-				out = errs[0].Error
-			} else {
-				out = fmt.Sprintf("%#v", validationError.DetailedOutput().Errors)
-			}
-
-			return fmt.Errorf("module %q argument %q validation failed: %s", s.t.Module.Name, pth, out)
+			return fmt.Errorf("module %q argument %q validation failed: %#v",
+				s.t.Module.Name, pth, validationError.DetailedOutput())
 		}
 
 		return errors.Wrapf(err, "module %q argument %q validation failed", s.t.Module.Name, pth)


internal/codegen/tpl_stencil_arg.go
```diff
@@ -205,17 +205,8 @@ func (s *TplStencil) validateArg(pth string, arg *configuration.Argument, v inte
...
-			// return the full list of errors.
-			errs := validationError.DetailedOutput().Errors
-			out := ""
-			if len(errs) == 1 {
-				out = errs[0].Error
-			} else {
-				out = fmt.Sprintf("%#v", validationError.DetailedOutput().Errors)
-			}
-
-			return fmt.Errorf("module %q argument %q validation failed: %s", s.t.Module.Name, pth, out)
```
artemchernyak-outreach, line 218: I really like the spirit of this change. I think having the full path is very helpful, but I am worried that this will be too much information and overwhelm the user.

Instead of using the `pth` here, could we make a nice output using a combination of `AbsoluteKeywordLocation` and `InstanceLocation`?

samnguyen-outreach: Cool, yeah this PR captures the spirit without any polish.

There is a tricky detail here which is that `DetailedOutput` is recursive and I hadn't quite decided how to handle that, but I suppose we can just provide some recursive renderer that just captures the two fields you mentioned.  What do you think of that approach?

jaredallard: I think there's a better way for us to represent this for sure. I'd also like to print out this information instead of returning it in the error signature, that would help us represent this better.

I'm open to giving you an idea of how to do it, or I can take it on myself here real quick. Let me know! 

Thanks so much for the contribution and conversation starter though :heart:

samnguyen-outreach: Please feel free to take it, I’m happy to just be the conversation starter here. Thanks Artem and Jared.

samnguyen-outreach:
I’m going to close the proposal PR for now since I’m not actively working on it.  Feel free to resurrect and take it over.

PR #174, opened 2022-11-14
fix: support from manifest arugments in tests

artemchernyak-outreach:

Support from was a matter of loading dependent modules in tests.

You can see this working in https://github.com/getoutreach/stencil-outreach/pull/59.

## Jira ID

[DT-2896]

## Notes for your reviewers

[DT-2896]: https://outreach-io.atlassian.net/browse/DT-2896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/stenciltest/stenciltest.go
+++ b/pkg/stenciltest/stenciltest.go
@@ -113,7 +113,7 @@ func (t *Template) ErrorContains(msg string) {
 	t.errStr = msg
 }
 
-// getTemplateRepositoryNames retrieves the naems from the passed in TemplateRepositories
+// getTemplateRepositoryNames retrieves the names from the passed in TemplateRepositories
 func getTemplateRepositoryNames(trs []*configuration.TemplateRepository) []string {
 	deps := make([]string, len(trs))
 	for i, tr := range trs {
--- a/pkg/stenciltest/stenciltest.go
+++ b/pkg/stenciltest/stenciltest.go
@@ -8,6 +8,7 @@ package stenciltest
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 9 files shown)


pkg/stenciltest/stenciltest.go
```diff
@@ -112,17 +113,55 @@ func (t *Template) ErrorContains(msg string) {
...
+	return deps
+}
+
+// modulesFromTemplateRepository creates new modules from the passed in TemplateRepositories
+func modulesFromTemplateRepository(trs []*configuration.TemplateRepository) ([]*modules.Module, error) {
+	var mods []*modules.Module
+	for _, tr := range trs {
+		// Use the present version or main
+		if tr.Version == "" {
+			tr.Version = "main"
```
jaredallard, line 131: I think if `tr.Version` is left undefined it'll determine the latest version to use... I don't like assuming the default branch here

pkg/stenciltest/stenciltest.go
```diff
@@ -112,17 +113,55 @@ func (t *Template) ErrorContains(msg string) {
 	t.errStr = msg
 }
 
+// getTemplateRepositoryNames retrieves the naems from the passed in TemplateRepositories
```
jaredallard, line 116:
```suggestion
// getTemplateRepositoryNames retrieves the names from the passed in TemplateRepositories
```

pkg/stenciltest/stenciltest.go
artemchernyak-outreach: It doesn't automatically determine it. Let me see if I can refactor some of the version resolution code so it can be used here.

artemchernyak-outreach:
`stencil-outreach` has been updated to use the latest commit from this branch, and all tests still pass: https://github.com/getoutreach/stencil-outreach/pull/59.

jaredallard (approved):

PR #176, opened 2022-11-15
feat(modules): support resolving dependencies of in-memory modules

jaredallard:

This PR adds support for resolving the dependencies of modules in-memory. This adds the notion of `dontResolve` to a `rm` (`ResolvedModule`)  that prevents a module from ever attempting to be resolved again (we re-resolve modules to ensure that additive criterias are considered).

This is used in combination with the new `ModuleResolveOptions` type that is passed to `GetModulesForService` that supports setting `Replacements` to be used to like the current `replacements` option in a `service.yaml`, but for in-memory modules.

## Jira ID

[XX-XX]

### DIFF
--- a/pkg/stenciltest/stenciltest.go
+++ b/pkg/stenciltest/stenciltest.go
@@ -16,6 +16,7 @@ import (
 	"testing"
 
 	"github.com/bradleyjkemp/cupaloy"
+	"github.com/getoutreach/gobox/pkg/cli/github"
 	"github.com/getoutreach/stencil/internal/codegen"
 	"github.com/getoutreach/stencil/internal/modules"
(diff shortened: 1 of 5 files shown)


pkg/stenciltest/stenciltest.go
```diff
@@ -75,12 +79,16 @@ func New(t *testing.T, templatePath string, additionalTemplates ...string) *Temp
 		t.Fatal(err)
 	}
 
+	log := logrus.New()
+	log.SetLevel(logrus.DebugLevel)
```
jaredallard, line 83: Figured it'd be useful to alway run tests w/ the debug logger to make it easier for people to identify issues

PR #182, opened 2022-11-28
fix: re-stencil 11-28-2022

artemchernyak-outreach:

Switches to stable version of all module and re-stencils. Test of full release.

## Jira ID

[XX-XX]


chatchai-outreach (approved):
restencil

PR #183, opened 2022-12-07
feat: globals

george-e-shaw-iv:
* go templates limit the scope of created variables to the template
  file they're created in. this sucks because sometimes you need to
  do complex argument parsing and then use the outputs of that
  parsing in multiple files.
* this allows you to do that^

## Jira ID

[XX-XX]

### DIFF
--- a/internal/codegen/tpl_stencil.go
+++ b/internal/codegen/tpl_stencil.go
@@ -53,6 +53,62 @@ func (s *TplStencil) GetModuleHook(name string) []interface{} {
 	return v
 }
 
+// SetGlobal sets a global to be used in the context of the current template module
+// repository. This is useful because sometimes you want to define variables inside
+// of a helpers template file after doing manifest argument processing and then use
+// them within one or more template files to be rendered; however, go templates limit
+// the scope of symbols to the current template they are defined in, so this is not
+// possible without external tooling like this function.
+//
+// This template function stores (and its inverse, GetGlobal, retrieves) data that is
+// not strongly typed, so use this at your own risk and be averse to panics that could
+// occur if you're using the data it returns in the wrong way.
+func (s *TplStencil) SetGlobal(name string, data interface{}) error {
+	// Only modify on first pass
+	if !s.s.isFirstPass {
+		return nil
+	}
+
+	// key is <module(self)>/<name>
+	k := path.Join(s.t.Module.Name, name)
+	s.log.WithField("template", s.t.ImportPath()).WithField("path", k).
+		WithField("data", spew.Sdump(data)).Debug("adding to global store")
+
+	v := reflect.ValueOf(data)
+	if !v.IsValid() {
+		err := fmt.Errorf("third parameter, data, must be set")
+		return err
+	}
+
+	// We just use the same map that AddToModuleHook uses, so we put it into a []interface{}
+	// even though we'll never actually use the data like that.
+	s.s.sharedData[k] = []interface{}{data}
+
+	return nil
+}
+
+// GetGlobal retrieves a global variable set by SetGlobal. The data returned from this function
+// is unstructured so by averse to panics - look at where it was set to ensure you're dealing
+// with the proper type of data that you think it is.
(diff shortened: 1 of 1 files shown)


george-e-shaw-iv:
waiting on jared before doing any more on this PR. there are no tests for other functions here other than readblocks. not sure if thats on purpose for some really weird reason or not?

internal/codegen/tpl_stencil.go
```diff
@@ -53,6 +53,62 @@ func (s *TplStencil) GetModuleHook(name string) []interface{} {
...
+// SetGlobal sets a global to be used in the context of the current template module
+// repository. This is useful because sometimes you want to define variables inside
+// of a helpers template file after doing manifest argument processing and then use
+// them within one or more template files to be rendered; however, go templates limit
+// the scope of symbols to the current template they are defined in, so this is not
+// possible without external tooling like this function.
+//
+// This template function stores (and its inverse, GetGlobal, retrieves) data that is
+// not strongly typed, so use this at your own risk and be averse to panics that could
+// occur if you're using the data it returns in the wrong way.
```
jaredallard, line 65: nit: add examples here in the function comment for the docs

internal/codegen/tpl_stencil.go
```diff
@@ -53,6 +53,62 @@ func (s *TplStencil) GetModuleHook(name string) []interface{} {
...
+	return nil
+}
+
+// GetGlobal retrieves a global variable set by SetGlobal. The data returned from this function
+// is unstructured so by averse to panics - look at where it was set to ensure you're dealing
+// with the proper type of data that you think it is.
+func (s *TplStencil) GetGlobal(name string) interface{} {
+	k := path.Join(s.t.Module.Name, name)
+
+	if v, ok := s.s.sharedData[k]; ok {
```
jaredallard, line 96: I'd prefer we still handle `v` not having an index, just in case

internal/codegen/tpl_stencil.go
```diff
@@ -53,6 +53,62 @@ func (s *TplStencil) GetModuleHook(name string) []interface{} {
...
+
+	v := reflect.ValueOf(data)
+	if !v.IsValid() {
+		err := fmt.Errorf("third parameter, data, must be set")
+		return err
+	}
+
+	// We just use the same map that AddToModuleHook uses, so we put it into a []interface{}
+	// even though we'll never actually use the data like that.
+	s.s.sharedData[k] = []interface{}{data}
```
jaredallard, line 85: Talked offline, we want to change sharedData to be a custom type that has Globals and ModuleHooks on it.

internal/codegen/tpl_stencil.go
george-e-shaw-iv: v not having an index? what

internal/codegen/tpl_stencil.go
```diff
@@ -45,14 +44,75 @@ type TplStencil struct {
...
+// them within one or more template files to be rendered; however, go templates limit
+// the scope of symbols to the current template they are defined in, so this is not
+// possible without external tooling like this function.
+//
+// This template function stores (and its inverse, GetGlobal, retrieves) data that is
+// not strongly typed, so use this at your own risk and be averse to panics that could
+// occur if you're using the data it returns in the wrong way.
+//
+//	{{- /* This writes a global into the current context of the template module repository */}}
+//	{{ stencil.SetGlobal "IsGeorgeCool" true }}
```
jaredallard, line 67:
```suggestion
//	{{ stencil.SetGlobal "IsGeorgeCool" false }}
```

jaredallard (approved):

jaredallard:
I wonder if it'd be useful to store a container in each of these containing the file name (template) that wrote the value as well, thoughts? That should be accessible already.

internal/codegen/tpl_stencil.go
george-e-shaw-iv: lying in the codebase like that would be unethical jared

jaredallard (approved):

PR #192, opened 2023-01-03
chore: Promote unstable to rc

chatchai-outreach:
Promote unstable to rc, created by `dt release promote`

**DO NOT MERGE**


chatchai-outreach:
> @chatchai-outreach @jaredallard I think we need to hold off on releasing `stencil` in particular. The version resolver is still not working quite right, and we need to fix that before we run our next release.

sure closing it then

PR #200, opened 2023-01-13
feat: cache stencil modules using commit

artemchernyak-outreach:

Uses the commit for caching instead of the tag. This helps ensure we grab the same version even when we change the tag. Most commonly this happens with `unstable` as it is a dynamic tag.

## Jira ID

[DT-3191]

## Notes for your reviewers

[DT-3191]: https://outreach-io.atlassian.net/browse/DT-3191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/modules/module.go
+++ b/internal/modules/module.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/Masterminds/sprig/v3"
 	"github.com/getoutreach/gobox/pkg/cli/github"
+	"github.com/getoutreach/gobox/pkg/cli/updater/resolver"
 	"github.com/getoutreach/stencil/pkg/configuration"
 	"github.com/getoutreach/stencil/pkg/extensions"
 	"github.com/go-git/go-billy/v5"
@@ -121,7 +122,10 @@ func (m *Module) RegisterExtensions(ctx context.Context, log logrus.FieldLogger,
 		return nil
 	}
 
-	return ext.RegisterExtension(ctx, m.URI, m.Name, m.Version)
+	version := &resolver.Version{
+		Tag: m.Version,
+	}
+	return ext.RegisterExtension(ctx, m.URI, m.Name, version)
 }
 
 // Manifest downloads the module if not already downloaded and returns a parsed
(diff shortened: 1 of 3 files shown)


internal/modules/module.go
```diff
@@ -121,7 +122,10 @@ func (m *Module) RegisterExtensions(ctx context.Context, log logrus.FieldLogger,
 		return nil
 	}
 
-	return ext.RegisterExtension(ctx, m.URI, m.Name, m.Version)
+	version := &resolver.Version{
```
jaredallard, line 125: I'd suggest using the function provided by gobox instead, it's meant to handle cases like this: https://pkg.go.dev/github.com/getoutreach/gobox/pkg/cli/updater/resolver#NewVersionFromVersionString

pkg/extensions/extensions.go
```diff
@@ -148,41 +150,79 @@ func (h *Host) RegisterInprocExtension(name string, ext apiv1.Implementation) {
...
+			return "", errors.Wrap(err, "retrieving commit")
+
+		}
+		version = v
 	}
 
 	// Check if the version we're pulling already exists and is executable before downloading
 	// it again.
-	dlPath := h.getExtensionPath(version, name)
+	dlPath := h.getExtensionPath(version.Commit, name)
```
jaredallard, line 225: I'd prefer to not lose the struct here, I'd recommend continuing to pass the version and handle the logic in `getExtensionPath` instead. It should be easier to test that way as well

internal/modules/module.go
jaredallard: Oh right, I thought that this might be the psudeo-version string but I guess it's not nvm!

pkg/extensions/extensions.go
```diff
@@ -148,36 +150,74 @@ func (h *Host) RegisterInprocExtension(name string, ext apiv1.Implementation) {
...
+		v, err = resolver.Resolve(ctx, token, &resolver.Criteria{
+			URL:     repoURL,
+			Channel: version.Tag,
+		})
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to get latest version")
+		}
+		return v, nil
+
+	}
```
jaredallard, line 177: nit: :P 

```suggestion
	}

```

jaredallard (approved):

PR #201, opened 2023-01-18
feat: add module configure command

kaldorn:
Adds a `stencil update module` command to either add or remove the `native-extension` fields in the service.yaml file.

## Jira ID

[DT-2972]

## Notes for your reviewers

[DT-2972]: https://outreach-io.atlassian.net/browse/DT-2972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/stencil/update_module.go
```diff
@@ -0,0 +1,84 @@
...
+			cmd := exec.CommandContext(c.Context, os.Args[0])
+			cmd.Stdout = os.Stdout
+			cmd.Stderr = os.Stderr
+			cmd.Stdin = os.Stdin
+			return errors.Wrap(cmd.Run(), "failed to run stancil")
+		},
+	}
+}
+
+func readAndMergeServiceYaml(path string, nativeExtension bool) error {
```
jaredallard, line 42: nit: function comment

cmd/stencil/update_module.go
```diff
@@ -0,0 +1,84 @@
...
+		Flags: []cli.Flag{
+			&cli.BoolFlag{
+				Name:  "native-extension",
+				Usage: "Updates a module to be a native extension",
+			},
+		},
+		Action: func(c *cli.Context) error {
+			readAndMergeServiceYaml("service.yaml", c.Bool("native-extension"))
+			//nolint:gosec // Why: intentional
+			cmd := exec.CommandContext(c.Context, os.Args[0])
```
jaredallard, line 33: nit: I'd suggest adding a comment about what you're doing here. Took me a second to realize you're trying to call yourself

cmd/stencil/update_module.go
```diff
@@ -0,0 +1,84 @@
...
+		Description: "updates a module with the provided name in the current directory",
+		ArgsUsage:   "update module",
+		Flags: []cli.Flag{
+			&cli.BoolFlag{
+				Name:  "native-extension",
+				Usage: "Updates a module to be a native extension",
+			},
+		},
+		Action: func(c *cli.Context) error {
+			readAndMergeServiceYaml("service.yaml", c.Bool("native-extension"))
```
jaredallard, line 31: Handle the error? :D

cmd/stencil/update_module.go
```diff
@@ -0,0 +1,84 @@
...
+
+	b, err := os.ReadFile(path)
+	if err != nil {
+		return err
+	}
+
+	err = yaml.Unmarshal(b, tm)
+	if err != nil {
+		return err
+	}
```
jaredallard, line 60: nit(non-blocking): Always handle errors in a single line, leads to less error shadowing issues

```suggestion
	if err := yaml.Unmarshal(b, tm); err != nil {
		return err
	}
```

cmd/stencil/update_module.go
```diff
@@ -0,0 +1,84 @@
...
+		delete(tm.Arguments, "plugin")
+	}
+	tm.Arguments["releaseOptions"] = releaseOpts
+
+	out, err := yaml.Marshal(tm)
+	if err != nil {
+		return err
+	}
+
+	err = os.WriteFile(path, out, 0o600)
```
jaredallard, line 79: Same here for the error handling

cmd/stencil/update_module.go
```diff
@@ -0,0 +1,84 @@
...
+	err = yaml.Unmarshal(b, tm)
+	if err != nil {
+		return err
+	}
+
+	releaseOpts := map[string]interface{}{
+		"enablePrereleases": true,
+	}
+
+	if nativeExtension {
```
jaredallard, line 66: Part of me feels like we should note what we're doing to the user, otherwise this might be somewhat confusing. Imagine a scenario where a user runs this, but forgets that they need to provide `--native-extension`. Maybe we should error if they don't provide a flag saying they would like to remove the native extension? Or ask for confirmation.

I don't have anything concrete here, but that's my thoughts right now.

kaldorn: That makes sense. Honestly, I think it's more useful to just have this default to adding the native-extension fields (based on the subcommand text) and have the flag be to remove them.

cmd/stencil/update.go
```diff
@@ -0,0 +1,21 @@
+// Copyright 2022 Outreach Corporation. All Rights Reserved.
+
+// Description: This file contains code for the create command
```
jaredallard, line 3: nit: file comment

cmd/stencil/update_module.go
```diff
@@ -0,0 +1,87 @@
...
+// update module command
+func NewUpdateModule() *cli.Command {
+	return &cli.Command{
+		Name:        "module",
+		Description: "updates a module with the provided name in the current directory",
+		ArgsUsage:   "update module",
+		Flags: []cli.Flag{
+			&cli.BoolFlag{
+				Name:  "--remove-native-extension",
+				Usage: "Updates a module to be a native extension",
```
jaredallard, line 27: Usage string seems to be wrong, should it be "Removes native extension configuration for the provided module"?

cmd/stencil/update_module.go
```diff
@@ -0,0 +1,87 @@
...
+
+	if err := yaml.Unmarshal(b, tm); err != nil {
+		return err
+	}
+
+	releaseOpts := map[string]interface{}{
+		"enablePrereleases": true,
+	}
+
+	if !removeNativeExtension {
```
jaredallard, line 70: Do we want to error if the user attempts to remove a native extension from a module that doesn't have one?

cmd/stencil/update_module.go
```diff
@@ -0,0 +1,87 @@
...
+		tm.Arguments["plugin"] = true
+		releaseOpts["force"] = true
+	} else {
+		delete(tm.Arguments, "plugin")
+	}
+	tm.Arguments["releaseOptions"] = releaseOpts
+
+	out, err := yaml.Marshal(tm)
```
jaredallard, line 78: We should warn/ask for confirmation about overwritting the user's `service.yaml` since this will strip all comments (unless you want to do the work to parse this into a `yaml.Node`

cmd/stencil/update.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"github.com/urfave/cli/v2"
+)
+
+// NewCreateCommand returns a new urfave/cli.Command for the
+// create command
+func NewUpdateCommand() *cli.Command {
+	return &cli.Command{
+		Name:        "update",
```
jaredallard, line 15: I don't have a better idea, but `update` seems... a little too loose/generic here. What this command really does is configure a module. Maybe it should be `module configure`?

cmd/stencil/configure.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"github.com/urfave/cli/v2"
+)
+
+// NewCreateCommand returns a new urfave/cli.Command for the
+// create command
+func NewConfigureCommand() *cli.Command {
+	return &cli.Command{
+		Name:        "configure",
+		Description: "Commands to configure template repositories for native-extension functionality, or stencil powered repositories",
```
jaredallard, line 16: I think we want the inverse, right? `module configure`? Not `configure module`

kaldorn: Makes sense.

cmd/stencil/configure_module.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/getoutreach/stencil/pkg/configuration"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli/v2"
+	"gopkg.in/yaml.v3"
+)
+
+// NewConfigureModule returns a new urfave/cli.Command for the
+// update module command
+func NewConfigureModule() *cli.Command {
```
jaredallard, line 22:
```suggestion
func NewConfigureModuleCmd() *cli.Command {
```

jaredallard: Or `Command`

jaredallard (approved):

PR #229, opened 2023-03-15
fix: panic on GetModuleHook that doesnt exist

jaredallard:

This PR fixes a panic on a module hook that has no data.

## Jira ID

[XX-XX]

### DIFF
--- a/internal/codegen/tpl_stencil.go
+++ b/internal/codegen/tpl_stencil.go
@@ -36,19 +36,19 @@ type TplStencil struct {
 // This is incredibly useful for allowing other modules to write
 // to files that your module owns. Think of them as extension points
 // for your module. The value returned by this function is always a
-// []interface{}, aka a list.
+// []any, aka a list.
 //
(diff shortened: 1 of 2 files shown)


internal/codegen/tpl_stencil.go
```diff
@@ -46,6 +46,10 @@ type TplStencil struct {
 func (s *TplStencil) GetModuleHook(name string) []interface{} {
 	k := s.s.sharedData.key(s.t.Module.Name, name)
 	v := s.s.sharedData.moduleHooks[k]
+	if v == nil {
+		// No data, return nothing
+		return nil
```
deregtd, line 51: Per your docs, shouldn't this be `return make([]interface, 0)` ?

```
This is incredibly useful for allowing other modules to write
// to files that your module owns. Think of them as extension points
// for your module. The value returned by this function is always a
// []interface{}, aka a list.
```

deregtd (approved):
returns empty array if there's no module data for the asked hook in GetModuleHook

PR #270, opened 2023-08-15
fix: Use OrbVersion for correct shared orb version

yj-yan:
…b version

Add OrbVersion to get the shared orb version and pass to stencil-circleCI. This is used to fix the problem that stencil-circleci changed the shared orb to `devbase:local` in devbase repo.

## Jira ID

[DT-3834]

## Notes for your reviewers
The OrbVersion is only applied when it is set in the manifest.

### DIFF
--- a/internal/codegen/values.go
+++ b/internal/codegen/values.go
@@ -79,6 +79,9 @@ type module struct {
 
 	// Version is the version of the current module
 	Version string
+
+	// OrbVersion is the shared orb version for current module
+	OrbVersion string
 }
 
 // stencilTemplate contains information about the current template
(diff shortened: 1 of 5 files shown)


jaredallard (changes requested):
Unfortunately, I don't think this PR makes a ton of sense as-is right now. If we set `OrbVersion` here, we're opening ourselves up to project-specific version information that doesn't really have anything to do with stencil modules (what happens when someone wants a custom docker image version, or some other artifact?).

I'd propose that we instead encapsulate any information that needs to be specific to a project into a [native module](https://engineering.outreach.io/stencil/reference/native-extensions/) instead, which is what was meant to introduce functionality like this!

yj-yan:
> Unfortunately, I don't think this PR makes a ton of sense as-is right now. If we set `OrbVersion` here, we're opening ourselves up to project-specific version information that doesn't really have anything to do with stencil modules (what happens when someone wants a custom docker image version, or some other artifact?).
> 
> I'd propose that we instead encapsulate any information that needs to be specific to a project into a [native module](https://engineering.outreach.io/stencil/reference/native-extensions/) instead, which is what was meant to introduce functionality like this!


Thanks for suggestion. I see the risk here... I will look into the way of using native module instead for implementation. Closing this PR for now.

PR #278, opened 2023-09-06
feat(stenciltest): add support for setting modules

malept:

This allows a test to add one or more Modules to a test service manifest, so that it can be used when rendering a snapshot.

[DT-4734]

[DT-4734]: https://outreach-io.atlassian.net/browse/DT-4734?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jaredallard (commented):
Hmmm, I'm not sure I like this change. It seems like this would open us up to allowing arbitrary modules being added in tests, which could no longer be indicative of what is being used at runtime (therefore reducing the usefulness of the test).

Perhaps we should just change modules to have a `manifest.lock`, or something along those lines, that lock the versions being used for tests?

malept:
I think all I really want is a way to specify that an a loaded module dependency's version is different from the current stable version (in the example I'm working off of, I want `devbase` to be the `main` version). Which sounds like what you're proposing, although I'd like to see what that API looks like.

malept:
I have a workaround for this so I don't need it anymore. At some point it would be nice to mock modules, though.

malept:
This is annoying me enough (multiple failed releases of `stencil-circleci`) that I'm reopening this.

I did talk to Jared about this shortly before I closed this PR initially, and he was less opposed to the changes.

Additionally, Stencil v2 does not currently have a test framework, so any changes to `stenciltest` won't matter in the long run, as *all* tests will have to be reworked whenever upstream adds a test framework.

ekalinichev (approved):

PR #315, opened 2024-03-08
feat(stenciltest): add RegenerateSnapshots function

malept:

Makes it easier to regenerate snapshots locally, but also show a diff (and not regenerate them) in CI.


yj-yan (approved):

kaldorn (approved):
This is good.

PR #338, opened 2024-05-21
feat: parallel resolution

AndrewWinterman:
Stencil spends a little while just doing network requests to gather its
dependencies.

Anecdotally, this is like several seconds per module, which makes
resolution take a long time.

This change set adds a smidge of parallelism to that resolution. The
strategy is to launch a fixed number of worker routines, use them to
resolve the work list, and then on completion, check whether items were
added to the work list that no worker picked up, due to a race condition
between when workers check for more work and when work can get added.

Existing tests covered this changeset completely, including revealing
situations where we'd leave items in the work set, so I didn't feel the
need to add additional coverage.

### DIFF
--- a/internal/modules/modules.go
+++ b/internal/modules/modules.go
@@ -12,6 +12,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/getoutreach/gobox/pkg/cfg"
@@ -19,6 +20,7 @@ import (
 	"github.com/getoutreach/stencil/pkg/configuration"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/errgroup"
 )
 
 // resolvedModule is used to keep track of a module during the resolution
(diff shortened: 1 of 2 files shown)


internal/modules/modules.go
```diff
@@ -96,16 +98,214 @@ type ModuleResolveOptions struct {
...
+
+	return &workItem{importPath: importPath, inProgressResolution: rm, spec: resolv, uri: uri}
+}
+
+// push adds a module to the work list
+func (list *workList) push(task *resolveModule) {
+	list.mu.Lock()
+	defer list.mu.Unlock()
+
+	list.tasks = append(list.tasks, *task)
```
george-e-shaw-iv, line 300: why deref as opposed to just queuing the pointers?

internal/modules/modules.go
```diff
@@ -139,154 +339,33 @@ func GetModulesForService(ctx context.Context, opts *ModuleResolveOptions) ([]*M
...
-	m *resolveModule, resolved map[string]*resolvedModule) (*resolver.Version, error) {
+func (list *workList) getLatestModuleForConstraints(ctx context.Context, item *workItem, token cfg.SecretData) (*resolver.Version, error) {
 	constraints := make([]string, 0)
-	for _, r := range resolved[m.conf.Name].history {
+	m := item.spec
+
+	list.mu.Lock()
+	// todo(AWinterman): Did i fuckover some modify by reference by lifting this up
+	module, ok := list.resolved[m.conf.Name]
+	// todo(AWinterman): this looks the same as item.inProgressResolution, is it?
```
george-e-shaw-iv, line 358: Are these still valid comments?

george-e-shaw-iv: I also don't see where you check the value of `ok`

internal/modules/modules.go
```diff
@@ -96,16 +98,214 @@ type ModuleResolveOptions struct {
 //
 //nolint:funlen // Why: Will be refactored in the future
 func GetModulesForService(ctx context.Context, opts *ModuleResolveOptions) ([]*Module, error) {
-	// start resolving the top-level modules
-	modulesToResolve := make([]resolveModule, 0)
+	wl := buildWorkLists(opts)
+
+	log := opts.Log
+
+	concurentResolutions := 5
```
george-e-shaw-iv, line 105: Should probably be flag driven with a default

george-e-shaw-iv (approved):
Looks good, no major comments

internal/modules/modules.go
AndrewWinterman: Could do. Aren't slices already pointers,

internal/modules/modules.go
```diff
@@ -96,16 +98,214 @@ type ModuleResolveOptions struct {
...
+			Mutable: true,
+		}
+
+		// if uri is local, represent it as "local" instead of the full path
+		if uriIsLocal(item.uri) {
+			version.Tag = "local"
+		} else {
+			// otherwise assume in-memory
+			version.Tag = "in-memory"
+		}
```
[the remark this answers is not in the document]
AndrewWinterman: Nice that is better. I just copied what Jared did. !blame

internal/modules/modules.go
george-e-shaw-iv: They're reference types, but that is just the container. You're still copying the entire `task` struct into the container (the slice) as opposed to just the reference here though

AndrewWinterman: true, i can make it a slice of pointers if you care. seemed low impact to me

internal/modules/modules.go
AndrewWinterman: fair

internal/modules/modules.go
AndrewWinterman: there's still some confusion iaround when to pull the module out of the map again vs use the argument. I use passive voice because that inconsistency was there before I showed up. I think I should expriment with simply referring to the argument, saves a lock.

AndrewWinterman: `ok` is referenced way down here: https://github.com/getoutreach/stencil/pull/338/files/dfdec3c4e1062510f92d9e928c8d6f879d112165#diff-9dc0e6e029819e3b3e02f006474df4060161209300437c5dfa2cc6a1a0815e8bR386

internal/modules/modules.go
```diff
@@ -139,154 +339,33 @@ func GetModulesForService(ctx context.Context, opts *ModuleResolveOptions) ([]*M
...
+
+	for _, r := range module.history {
 		if r.constraint != "" {
 			constraints = append(constraints, r.constraint)
 		}
 	}
 
 	channel := m.conf.Channel
-	for _, r := range resolved[m.conf.Name].history {
+	for _, r := range module.history {
```
george-e-shaw-iv, line 368: These two calls could panic since you're not checking `ok` until line 386

internal/modules/modules.go
AndrewWinterman: done.

internal/modules/modules.go
AndrewWinterman: i removed the comments, we still want to check the map ref i guess.

internal/modules/modules.go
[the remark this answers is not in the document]
AndrewWinterman: done

internal/modules/modules.go
AndrewWinterman: done

internal/modules/modules.go
AndrewWinterman: that is true, but as you can sort of see from the dif i just copied this from elsewhere.

AndrewWinterman: I think it's actually guaranteed to be in the map by externalities i don't quite understand... 

I can fix if you'd like to eliminate that possibility. Lemme know.

AndrewWinterman: okay, fixed.

cmd/stencil/stencil.go
```diff
@@ -85,14 +85,27 @@ func main() {
...
 			return errors.Wrap(cmd.Run(ctx), "run codegen")
 		},
 		// <</Stencil::Block>>
 	}
 	app.Flags = []cli.Flag{
 		// <<Stencil::Block(flags)>>
+		&cli.StringFlag{
+			Name:        "concurrent-resolvers",
+			Aliases:     []string{"c"},
+			DefaultText: "5",
```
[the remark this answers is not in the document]
AndrewWinterman: Yeah... I considered that but I think the complexity trade off might not quite be there, especially since you could reasonably have quite a few routines waiting on network without meaningfully impacting cpu at all

internal/modules/modules.go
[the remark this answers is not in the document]
jaredallard: rude @AndrewWinterman

jaredallard (approved):
Not that my approval means anything on here, but good idea. Never occurred to me to do parallelization of module resolution and `stencil` _def_ spent a lot of time doing that.

malept (approved):

PR #357, opened 2024-09-11
chore: maintenance mode announcement

malept:

This is for three reasons:

1. Note that a community version exists
2. Make it clear that we don't have the time to handle community requests
3. Allow the community version to use the Homebrew package identifier instead ([ref](https://github.com/orgs/Homebrew/discussions/5587))

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,9 @@
 
 <!-- <</Stencil::Block>> -->
 
+> [!WARNING]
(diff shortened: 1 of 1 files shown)


jaredallard:
Thanks @malept! If there's anything I can help with on my side, let me know y'all 🚀

Ned-el-ch (approved):

Ned-el-ch:
I approve this message

PR #389, opened 2025-03-04
feat: allow underscores in blocks

AndrewWinterman:
Please read [CONTRIBUTING.md](CONTRIBUTING.md) for additional information on contributing to this repository!

## Jira ID

[XX-XX]


malept (changes requested):
Blocking this until we see whether upstream stencil will take the same change. I want to avoid too much divergence in the event that we migrate to upstream, unless there's a really good reason to do so.

malept (approved):
This was added upstream (https://github.com/rgst-io/stencil/pull/249) so we can take this

jaredallard:
Note that we're deprecating v1 blocks. Unsure if y'all even use v1 blocks any more, but figured I'd call it out. It's likely we'll cut a V3 mid year or so that fully nukes it.

PR #396, opened 2025-03-13
fix: add failed command in error message for post-run commands

iamrz1:

Clarifies what command failed if a post-run command failed for a stencil module

## Jira ID

[DT-4609]

## Notes for your reviewers

[DT-4609]: https://outreach-io.atlassian.net/browse/DT-4609?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/codegen/stencil_test.go
```diff
@@ -121,3 +121,30 @@ func ExampleStencil_PostRun() {
 	// Output:
 	// hello
 }
+
+func TestStencilPostRunError(t *testing.T) {
+	ctx := context.Background()
+	m, err := modules.New(ctx, "", &configuration.TemplateRepository{Name: "github.com/getoutreach/stencil-base", Version: "main"})
+	if err != nil {
+		t.Errorf("failed to create module: %v", err)
+	}
```
malept, line 130: suggestion: use [`assert.NilError`](https://pkg.go.dev/gotest.tools/v3@v3.5.1/assert#NilError) instead.

internal/codegen/stencil_test.go
```diff
@@ -121,3 +121,30 @@ func ExampleStencil_PostRun() {
...
+
+	manifest, err := m.Manifest(ctx)
+	if err != nil {
+		t.Errorf("failed to get manifest from module: %v", err)
+	}
```
malept, line 135: suggestion: use [`assert.NilError`](https://pkg.go.dev/gotest.tools/v3@v3.5.1/assert#NilError) instead.

internal/codegen/stencil_test.go
```diff
@@ -121,3 +121,30 @@ func ExampleStencil_PostRun() {
...
+
+	if len(manifest.PostRunCommand) == 0 {
+		t.Errorf("manifest does not contain post-run commands")
+	}
```
malept, line 139: suggestion: use [`assert.Assert`](https://pkg.go.dev/gotest.tools/v3@v3.5.1/assert#Assert) instead.

internal/codegen/stencil_test.go
```diff
@@ -121,3 +121,30 @@ func ExampleStencil_PostRun() {
...
+	prc := manifest.PostRunCommand[0]
+
+	st := NewStencil(&configuration.ServiceManifest{
+		Name:      "TestStencilPostRunError",
+		Arguments: map[string]interface{}{},
+	}, []*modules.Module{m}, logrus.New())
+
+	err = st.PostRun(ctx, logrus.New())
+	exErr := fmt.Sprintf("failed to run post run command for module %s and command %s", m.Name, prc.Command)
```
malept, line 148: suggestion: the module and command should be hardcoded so it's more obvious what the intended behavior should be for future engineers trying to debug this test.

internal/codegen/stencil_test.go
```diff
@@ -121,3 +121,30 @@ func ExampleStencil_PostRun() {
...
+	}
+
+	if len(manifest.PostRunCommand) == 0 {
+		t.Errorf("manifest does not contain post-run commands")
+	}
+	prc := manifest.PostRunCommand[0]
+
+	st := NewStencil(&configuration.ServiceManifest{
+		Name:      "TestStencilPostRunError",
+		Arguments: map[string]interface{}{},
```
malept, line 144: Nit: `gopls` has been nagging me lately to use `any` instead of `interface{}` for readability purposes.

malept (changes requested):

iamrz1:
I reworked the test to load the manifest.yaml from a local file (using memfs) instead of the remote file from github. This will make the error reproduction independent of any changes to the upstream file. The command and module names are also hardcoded so it is more obvious to debug.

internal/codegen/stencil_test.go
```diff
@@ -121,3 +122,25 @@ func ExampleStencil_PostRun() {
 	// Output:
 	// hello
 }
+
+func TestStencilPostRunError(t *testing.T) {
+	var (
+		name    = "TestStencilPostRunError"
+		command = strings.ToLower("invalidPostRunCommand")
```
malept, line 129: question: why does this need to be all lowercase?

If it does need to be lowercase, why not just hardcode it that way and save the function call?

malept (changes requested):
Thank you for refactoring the code to not use an existing module, it's definitely preferred that unit tests should not make network calls unless absolutely necessary.

internal/codegen/stencil_test.go
iamrz1: it's for cosmetic purpose, commands usually come in lowercase, but `invalidpostruncommand` looks like a typo. Do you recommend `command = "invalidPostRunCommand"` or `command = "invalidpostruncommand" instead?

malept: It doesn't matter since it's a test value. Just use the first option.

internal/codegen/stencil_test.go
```diff
@@ -121,3 +121,25 @@ func ExampleStencil_PostRun() {
...
+		command = "invalidPostRunCommand"
+	)
+
+	ctx := context.Background()
+	fs := memfs.New()
+	mf, err := fs.Create("manifest.yaml")
+	assert.NilError(t, err)
+	_, err = fmt.Fprintf(mf, "name: %s\npostRunCommand:\n- command: %s\n", name, command)
+	assert.NilError(t, err)
+	assert.NilError(t, mf.Close())
```
malept, line 137: suggestion: can we refactor this to be a test helper since this pattern is in at least two other tests, and it makes sense for this to be used in future tests? Something like:

```go
func createFakeModuleWithManifest(
  t *testing.T,
  manifestContents string,
) billy.Filesystem
```

This also gives us the opportunity to `defer` the `mf.Close()` function call, although you maybe would still want to check the error, it depends on what errors `.Close()` gives.

malept (changes requested):

internal/codegen/stencil_test.go
iamrz1: Yes, we can create a separate method for this.
We don't need to keep the manifest file open, so we don't really have to defer the `mf.Close()`.

malept: > We don't need to keep the manifest file open, so we don't really have to defer the `mf.Close()`.

I think you'll want to `defer` the `.Close()` call in the event that `assert.NilError()` is triggered for your `.Write()` call.

internal/codegen/stencil_test.go
```diff
@@ -121,3 +122,34 @@ func ExampleStencil_PostRun() {
...
+	)
+
+	fs := createFakeModuleFSWithManifest(t, manifestContents)
+	st := NewStencil(&configuration.ServiceManifest{Name: name},
+		[]*modules.Module{modules.NewWithFS(ctx, name, fs)}, logger)
+
+	assert.ErrorContains(t, st.PostRun(ctx, logger), errMsg)
+}
+
+func createFakeModuleFSWithManifest(t *testing.T, manifestContents string) billy.Filesystem {
```
malept, line 146: suggestion: please add a doc comment for this function and also use it in the other two tests that write manifest files.

internal/codegen/stencil_test.go
```diff
@@ -121,3 +122,34 @@ func ExampleStencil_PostRun() {
...
+	var (
+		ctx              = context.Background()
+		logger           = logrus.New()
+		manifestContents = fmt.Sprintf("name: %s\npostRunCommand:\n- command: %s\n", name, command)
+		errMsg           = fmt.Sprintf("failed to run post run command for module %s and command %s", name, command)
+	)
+
+	fs := createFakeModuleFSWithManifest(t, manifestContents)
+	st := NewStencil(&configuration.ServiceManifest{Name: name},
+		[]*modules.Module{modules.NewWithFS(ctx, name, fs)}, logger)
```
malept, line 141: nit: this is personal preference but I think it reads better if you put each argument on their own line.
```suggestion
	st := NewStencil(
		&configuration.ServiceManifest{Name: name},
		[]*modules.Module{modules.NewWithFS(ctx, name, fs)},
		logger,
	)
```

malept (changes requested):

internal/codegen/stencil_test.go
```diff
@@ -121,3 +116,47 @@ func ExampleStencil_PostRun() {
...
+// is useful for testing purposes where a mock filesystem is needed.
+//
+// Parameters:
+//   - t: The testing object used for assertions.
+//   - manifestContents: A string representing the contents to be written to the manifest file.
+//
+// Returns:
+//   - A billy.Filesystem representing the in-memory filesystem with the manifest file.
+func createFakeModuleFSWithManifest(t *testing.T, manifestContents string) billy.Filesystem {
+	fs := memfs.New()
```
malept, line 154: suggestion: denote to the testing context that this is a [helper](https://pkg.go.dev/testing#T.Helper). This is so that when the file/line of the assertion error is printed out, it's not printing the information of the helper function.
```suggestion
func createFakeModuleFSWithManifest(t *testing.T, manifestContents string) billy.Filesystem {
	t.Helper()
	fs := memfs.New()
```

internal/codegen/stencil_test.go
```diff
@@ -121,3 +116,47 @@ func ExampleStencil_PostRun() {
...
+// file named "manifest.yaml" containing the provided manifest contents. This
+// is useful for testing purposes where a mock filesystem is needed.
+//
+// Parameters:
+//   - t: The testing object used for assertions.
+//   - manifestContents: A string representing the contents to be written to the manifest file.
+//
+// Returns:
+//   - A billy.Filesystem representing the in-memory filesystem with the manifest file.
+func createFakeModuleFSWithManifest(t *testing.T, manifestContents string) billy.Filesystem {
```
malept, line 153: suggestion: consider putting this at the top rather than the bottom of the file. Or perhaps in its own test helpers file, otherwise it might get lost if someone decides to put more tests at the bottom of the file.

internal/codegen/stencil_test.go
```diff
@@ -113,11 +104,29 @@ func ExampleStencil_PostRun() {
 	}, []*modules.Module{
 		modules.NewWithFS(ctx, "testing", fs),
 	}, logrus.New())
-	err := st.PostRun(ctx, nullLog)
-	if err != nil {
-		fmt.Println(err)
-	}
 
-	// Output:
-	// hello
```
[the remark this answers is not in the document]
iamrz1: Did not seem relevant to the test. The command prints `hello`, we don't need a comment to explain that.

internal/codegen/fake_helpers_test.go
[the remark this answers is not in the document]
iamrz1: okay, can do.

malept (changes requested):

internal/codegen/stencil_test.go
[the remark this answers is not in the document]
iamrz1: I can revert it back to an example test.

internal/codegen/stencil_test.go
```diff
@@ -95,14 +91,11 @@ func TestModuleHookRender(t *testing.T) {
 	assert.Equal(t, tpls[1].Files[0].String(), "a", "expected Render() to return correct output")
 }
 
-func ExampleStencil_PostRun() {
-	fs := memfs.New()
+func TestExampleStencil_PostRun(t *testing.T) {
```
malept, line 94: suggestion: I don't think that's how Examples are supposed to be defined, according to the docs. In this case you may just want to revert the changes to this function and leave the `memfs` code the way it was.

internal/codegen/test_helpers.go
malept, line 1: suggestion: I don't know why the linter didn't catch this, but this file is missing the copyright and file description header, like all of the other files.

malept (changes requested):

internal/codegen/test_helpers.go
iamrz1: Ah, yes! Will add it.

iamrz1: Probably linter is not configured for it?

PR #418, opened 2025-06-25
fix: only print debug log when debug log level is specified

malept:

Regression from #411


iamrz1 (approved):
I think I left it there by mistake.

PR #424, opened 2025-07-11
fix: --frozen-lockfile works with stable constraints and non-stable versions

malept:

See https://github.com/getoutreach/gobox/pull/701

This updates the gobox version to include the fix above, and adds a regression test.

## Jira ID

[DT-4795]

## Notes for reviewers

I validated that the test fails without the gobox change, and passes with the gobox change.

[DT-4795]: https://outreach-io.atlassian.net/browse/DT-4795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


malept:
Note: test failure is related to the fix for #423. If you run it locally (multiple times) the test passes.

ekalinichev (approved):

PR #445, opened 2025-11-11
fix: module replace not working

huazhoul-outreach:
The `replacements` is not working as expected. The root cause is the replacement configuration is only saved in `workList`.

This PR also add vscode configuration to debug stencil

Please read [CONTRIBUTING.md](CONTRIBUTING.md) for additional information on contributing to this repository!

## Jira ID

[XX-XX]

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -59,9 +59,19 @@
       "host": "127.0.0.1",
       "port": 42097
       // <</Stencil::Block>>
-    }
+    },
     // <<Stencil::Block(vscodeLaunchConfigs)>>
-
+    {
+      "name": "Debug stencil run",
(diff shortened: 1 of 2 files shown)


malept (changes requested):
It's unclear what problem this is solving. This PR needs:

* a clear set of steps to reproduce the problem
* one or more unit tests for future regression testing

I'm also worried about potential unforseen side effects from this change, but I haven't dug deeply into this part of the code to know exactly what those side effects could be.

FWIW, module replacement works for me and my team per [the docs](https://engineering.outreach.io/stencil/getting-started/module-quick-start/#step-3-consuming-the-module-in-an-application), specifically with relative paths.

PR #491, opened 2026-07-08
test(lint): snapshot-based TestValidate and TestFix

malept:

`TestValidate` and `TestFix` are the two largest manifest-lint table tests. They used custom matchers (`hasFinding`/`hasFindingMsg`, `want`/`wantMsg`/`none`, and per-case `wantOut`/`wantCount`) that are verbose and only check part of each result. This switches both to cupaloy snapshots of a rendered string, so each case is `{name, in}` plus a golden file.

cupaloy is already used in `pkg/stenciltest`, so there's no new dependency. This also adds a top-level `update-test-snapshots` mise task (`mise run update-test-snapshots`) to regenerate the goldens. It's kept out of the `generate:`/`test:` groups so a normal test run doesn't rewrite snapshots.

## Notes for your reviewers

- The snapshots assert more than the old tests. `TestValidate` previously checked severity and path (with some message substrings); the goldens now cover severity, path, line, and the full message. `TestFix` previously checked the fixed YAML and an applied count; the goldens now cover the YAML plus each applied change's path and message.
- `renderFindings` rewrites `file://…/manifest.yaml` to `file://manifest.yaml` before snapshotting. The JSON-schema compiler absolutizes that URL against the working directory, which would otherwise put the checkout path in the golden and break in CI. The substitution only touches that prefix; the `$ref` value (e.g. `file:///etc/hostname`) is left intact.
- Findings come out in a fixed check order with argument keys sorted, so the goldens are stable across machines. `TestValidateDeterministicOrder` stays as separate coverage of that ordering.
- Tests only: no production code under `internal/lint/**` changed, and the smaller sibling tests are untouched.

### DIFF
--- a/.mise/tasks/update-test-snapshots
+++ b/.mise/tasks/update-test-snapshots
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+#MISE description="Regenerate all cupaloy test snapshots (UPDATE_SNAPSHOTS=true go test ./...)."
+
+# Deliberately regenerates every committed .snapshots/ golden from current code.
+# Run this by hand after an intended behavior change, then review the diff before
+# committing. It must never be wired into a normal test run — auto-regenerating
+# snapshots would make them assert nothing.
+set -euo pipefail
+
+root="${MISE_PROJECT_ROOT:?MISE_PROJECT_ROOT is not set; run via 'mise run update-test-snapshots'}"
+
+echo "Regenerating cupaloy snapshots..."
+UPDATE_SNAPSHOTS=true go test -C "$root" ./...
--- a/.mise/tasks/update-test-snapshots
+++ b/.mise/tasks/update-test-snapshots
@@ -5,9 +5,30 @@
 # Run this by hand after an intended behavior change, then review the diff before
 # committing. It must never be wired into a normal test run — auto-regenerating
 # snapshots would make them assert nothing.
+#
+# With cupaloy's default FailOnUpdate, updating a snapshot makes `go test` exit
+# non-zero (each change is reported as a test error), so we can't rely on the
+# regeneration pass's exit code. Instead we regenerate, then run the tests again
+# normally: once the goldens match the code, that second pass fails only on real
+# problems (panics, compile errors, non-snapshot assertions).
+#
+# Note: pkg/stenciltest regenerates its template snapshots on a different trigger
+# (CI being unset, not UPDATE_SNAPSHOTS), so running this outside CI may also
+# refresh those. That RegenerateSnapshots() API is a kludge and future stencil
+# module testing will not carry it, so it is not a pattern to follow.
 set -euo pipefail
 
 root="${MISE_PROJECT_ROOT:?MISE_PROJECT_ROOT is not set; run via 'mise run update-test-snapshots'}"
 
 echo "Regenerating cupaloy snapshots..."
-UPDATE_SNAPSHOTS=true go test -C "$root" ./...
+UPDATE_SNAPSHOTS=true go test -C "$root" ./... >/dev/null 2>&1 || true
+
+echo "Verifying the suite passes against the regenerated snapshots..."
+if ! go test -C "$root" ./...; then
+  echo
+  echo "Tests still fail after regeneration — this is a real failure, not a snapshot update (see output above)." >&2
+  exit 1
+fi
+
+echo
+echo "Snapshots regenerated — review the diff before committing."
--- a/.mise/tasks/update-test-snapshots
+++ b/.mise/tasks/update-test-snapshots
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-#MISE description="Regenerate all cupaloy test snapshots (UPDATE_SNAPSHOTS=true go test ./...)."
+#MISE description="Regenerate and verify all cupaloy test snapshots"
 
 # Deliberately regenerates every committed .snapshots/ golden from current code.
 # Run this by hand after an intended behavior change, then review the diff before
--- a/internal/lint/manifest/manifest_test.go
+++ b/internal/lint/manifest/manifest_test.go
@@ -6,7 +6,6 @@ package manifest_test
 
 import (
 	"fmt"
-	"regexp"
 	"strings"
 	"testing"
 
@@ -16,12 +15,6 @@ import (
 	lint "github.com/getoutreach/stencil/internal/lint"
 	lintmanifest "github.com/getoutreach/stencil/internal/lint/manifest"
 )
-
-// absSchemaURL matches the machine-specific absolute file:// URL that the JSON
-// schema compiler embeds in error messages (it absolutizes the manifest URL
-// against the current working directory). Normalizing it keeps snapshots
-// portable across checkouts and CI without losing the meaningful suffix.
-var absSchemaURL = regexp.MustCompile(`file://[^ ]*/manifest\.yaml`)
 
 func TestLoadValid(t *testing.T) {
 	res, readErr := lintmanifest.Load(strings.NewReader("name: testing\n"))
@@ -74,12 +67,23 @@ func validateString(in string) []lint.Finding {
 	return lintmanifest.Validate(res)
 }
 
+// sevColWidth is the fixed width of the severity column in rendered findings.
+// It is len("warning"), the widest of the three severity values; widen this if
+// a longer severity is ever added.
+const sevColWidth = 7
+
 // renderFindings formats findings one per line as aligned columns
 // "SEVERITY  PATH:LINE  MESSAGE", or the literal "(no findings)" when empty,
 // for stable, readable snapshotting.
+//
+// It is used only for findings whose messages are stencil-owned and stable.
+// Cases whose message is third-party text (JSON-schema compiler output) or a
+// Go-internal type name (strict-decode errors) are asserted by severity+path in
+// TestValidateExternalErrors instead, so a dependency bump or type rename does
+// not churn a golden file.
 func renderFindings(findings []lint.Finding) string {
 	if len(findings) == 0 {
-		return "(no findings)"
+		return "(no findings)\n"
 	}
 	// Compute the path:line column width so the message column aligns.
 	locs := make([]string, len(findings))
@@ -92,8 +96,7 @@ func renderFindings(findings []lint.Finding) string {
 	}
 	var b strings.Builder
 	for i, f := range findings {
-		msg := absSchemaURL.ReplaceAllString(f.Message, "file://manifest.yaml")
-		fmt.Fprintf(&b, "%-7s  %-*s  %s\n", f.Severity, locWidth, locs[i], msg)
+		fmt.Fprintf(&b, "%-*s  %-*s  %s\n", sevColWidth, f.Severity, locWidth, locs[i], f.Message)
 	}
 	return b.String()
 }
@@ -113,10 +116,6 @@ func TestValidate(t *testing.T) {
 				"arguments:\n  greeting:\n    schema:\n      type: string\n",
 		},
 		{
-			name: "unknown top-level key",
-			in:   "name: testing\nnme: oops\n",
-		},
-		{
 			name: "missing name",
 			in:   "type: templates\n",
 		},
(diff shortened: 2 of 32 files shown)


.mise/tasks/update-test-snapshots
```diff
@@ -0,0 +1,13 @@
...
+# Deliberately regenerates every committed .snapshots/ golden from current code.
+# Run this by hand after an intended behavior change, then review the diff before
+# committing. It must never be wired into a normal test run — auto-regenerating
+# snapshots would make them assert nothing.
+set -euo pipefail
+
+root="${MISE_PROJECT_ROOT:?MISE_PROJECT_ROOT is not set; run via 'mise run update-test-snapshots'}"
+
+echo "Regenerating cupaloy snapshots..."
+UPDATE_SNAPSHOTS=true go test -C "$root" ./...
```
iamrz1: This success path exits non-zero. With the pinned cupaloy v2.3.0, the global `cupaloy.SnapshotT` runs with the default `FailOnUpdate(true)`, so regenerating a golden returns a `snapshot created/updated for test X` error that `SnapshotT` reports via `t.Error` — failing the test. As a result this command regenerates the goldens correctly but exits non-zero whenever any snapshot changes, and under `set -euo pipefail` the task aborts with no completion message.

Because `go test ./...` runs every package, a genuine failure (compile error, panic, non-snapshot assertion) is then indistinguishable from a normal regeneration. Capture the `go test` output/exit and, if the only failures are snapshot-update lines, print an explicit `snapshots regenerated — review the diff` and exit 0; otherwise surface the real error. Avoid a blanket `set +e`/exit-0 that would mask genuine failures. At minimum, document in the block comment that a non-zero exit with `snapshot updated` lines is the expected success signal.

internal/lint/manifest/manifest_test.go
```diff
@@ -65,185 +74,127 @@ func validateString(in string) []lint.Finding {
...
+	for i, f := range findings {
+		locs[i] = fmt.Sprintf("%s:%d", f.Path, f.Line)
+		if len(locs[i]) > locWidth {
+			locWidth = len(locs[i])
 		}
 	}
-	return false
+	var b strings.Builder
+	for i, f := range findings {
+		msg := absSchemaURL.ReplaceAllString(f.Message, "file://manifest.yaml")
```
iamrz1: This normalization strips only the machine-specific `file://` URL, but several goldens still pin volatile third-party error text and an internal Go type name that don't reflect stencil behavior:

- [`TestValidate-invalid_schema`](https://github.com/getoutreach/stencil/blob/cb86d4c8f672cbbef928637c295c098b86200fc2/internal/lint/manifest/.snapshots/TestValidate-invalid_schema#L1) pins the `santhosh-tekuri/jsonschema/v5` internal pointer `#/allOf/3/$ref/properties/type/anyOf/0/$ref/enum`.
- [`TestValidate-unknown_top-level_key`](https://github.com/getoutreach/stencil/blob/cb86d4c8f672cbbef928637c295c098b86200fc2/internal/lint/manifest/.snapshots/TestValidate-unknown_top-level_key#L1) and [`TestValidate-strict_failure_still_yields_field_findings`](https://github.com/getoutreach/stencil/blob/cb86d4c8f672cbbef928637c295c098b86200fc2/internal/lint/manifest/.snapshots/TestValidate-strict_failure_still_yields_field_findings#L1) pin `go.yaml.in/yaml/v3` wording plus the internal type name `configuration.TemplateRepositoryManifest`.

A patch bump of either dependency or a rename of `TemplateRepositoryManifest` would break these goldens with no change in stencil behavior — regressing the pre-refactor tests, which deliberately asserted only Path+Severity for these external-error cases. Consider extending the normalization here to cover the volatile library/type tail (as it already does for the URL), or retain a narrow Path+Severity assertion for the external-schema/strict-decode cases while snapshotting only the stencil-owned message text.

internal/lint/manifest/manifest_test.go
```diff
@@ -65,185 +74,127 @@ func validateString(in string) []lint.Finding {
...
+		fmt.Fprintf(&b, "%-7s  %-*s  %s\n", f.Severity, locWidth, locs[i], msg)
```
iamrz1: `%-7s` hardcodes `7` for the severity column while the adjacent `locWidth` is computed dynamically. The `7` silently equals `len("warning")`, the current max of the three-value severity enum, so it will misalign if a wider severity is ever added. Extract a named `const sevColWidth = 7` with a comment explaining the derivation, or compute it dynamically the same way `locWidth` is.

internal/lint/manifest/manifest_test.go
```diff
@@ -65,185 +74,127 @@ func validateString(in string) []lint.Finding {
...
-func hasFinding(findings []lint.Finding, sev lint.Severity, path, substr string) bool {
-	for _, f := range findings {
-		if f.Severity == sev && f.Path == path && strings.Contains(f.Message, substr) {
-			return true
+// renderFindings formats findings one per line as aligned columns
+// "SEVERITY  PATH:LINE  MESSAGE", or the literal "(no findings)" when empty,
+// for stable, readable snapshotting.
+func renderFindings(findings []lint.Finding) string {
+	if len(findings) == 0 {
+		return "(no findings)"
```
iamrz1: `renderFindings` returns `"(no findings)"` without a trailing newline, whereas its non-empty branch (and `renderFix`) ends every line with `\n`. Since cupaloy appends its own `\n`, the empty-case golden ends up shaped differently from the populated case. Align the empty case with the rest:

```suggestion
		return "(no findings)\n"
```

.mise/tasks/update-test-snapshots
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+#MISE description="Regenerate all cupaloy test snapshots (UPDATE_SNAPSHOTS=true go test ./...)."
+
+# Deliberately regenerates every committed .snapshots/ golden from current code.
+# Run this by hand after an intended behavior change, then review the diff before
+# committing. It must never be wired into a normal test run — auto-regenerating
+# snapshots would make them assert nothing.
```
iamrz1, line 7: Two snapshot-regeneration conventions now coexist: these new tests use the global `cupaloy.SnapshotT` gated on `UPDATE_SNAPSHOTS`, while [`pkg/stenciltest/stenciltest.go`](https://github.com/getoutreach/stencil/blob/cb86d4c8f672cbbef928637c295c098b86200fc2/pkg/stenciltest/stenciltest.go#L222) uses `RegenerateSnapshots()` gated on `os.Getenv("CI") == ""`. The two mechanisms key off different environment variables, which is easy to trip over. Consider either unifying on one trigger, or documenting both conventions here in the task comment so future contributors know which snapshots regenerate under which condition.

malept: Documented both triggers in the task comment; `RegenerateSnapshots()` is a kludge that future stencil module testing will drop, so the lint snapshots stay on `UPDATE_SNAPSHOTS` rather than following it.

.mise/tasks/update-test-snapshots
malept: Reworked the task to regenerate and then verify with a second normal `go test` pass, so real failures are no longer masked by the update exit code.

internal/lint/manifest/manifest_test.go
malept: Moved the external-error cases to a new `TestValidateExternalErrors` asserting severity and path only, and dropped the now-unused `absSchemaURL` normalization.

internal/lint/manifest/manifest_test.go
malept: Extracted `const sevColWidth = 7`, documented as `len("warning")`.

internal/lint/manifest/manifest_test.go
malept: The empty case now returns `"(no findings)\n"`, and the affected goldens were regenerated.

malept (commented):
:robot: _Thread responses generated by AI_

internal/lint/manifest/manifest_test.go
```diff
@@ -65,184 +67,171 @@ func validateString(in string) []lint.Finding {
...
-			// no error findings beyond those expected (warnings may include extras only if expected)
 			assert.Equal(t, len(test.want), len(got),
-				"finding count mismatch: want %d, got %v", len(test.want), got)
-			for path, substr := range test.wantMsg {
-				assert.Assert(t, hasFindingMsg(got, path, substr),
-					"missing message %q at %q in %v", substr, path, got)
+				"finding count mismatch; got %v", got)
+			for i, w := range test.want {
+				assert.Equal(t, w.Severity, got[i].Severity)
+				assert.Equal(t, w.Path, got[i].Path)
```
iamrz1, line 237: `TestValidateExternalErrors` now uses positional `got[i]` comparisons instead of the order-independent `hasFinding` set check, which silently tightens the finding-ordering contract. Order is code-guaranteed (`checkArguments` sorts via `sort.Strings`, and `TestValidateDeterministicOrder` locks it in), so flake risk is low — but the tightened contract is undocumented. Consider a one-line comment noting that finding order is a guaranteed, sorted contract, so a future reader understands the positional match is intentional rather than incidental.

internal/lint/manifest/manifest_test.go
```diff
@@ -65,184 +67,171 @@ func validateString(in string) []lint.Finding {
...
-// hasFinding reports whether findings contains one with the given severity and path
-// whose message contains substr.
-func hasFinding(findings []lint.Finding, sev lint.Severity, path, substr string) bool {
-	for _, f := range findings {
-		if f.Severity == sev && f.Path == path && strings.Contains(f.Message, substr) {
-			return true
+// sevColWidth is the fixed width of the severity column in rendered findings.
+// It is len("warning"), the widest of the three severity values; widen this if
+// a longer severity is ever added.
+const sevColWidth = 7
```
iamrz1: `sevColWidth = 7` is hand-derived from `len("warning")`, while `locWidth` just below is computed dynamically in the same function. The comment acknowledges the invariant ("widen this if a longer severity is ever added") but nothing enforces it — a new longer severity value would silently misalign every golden. Since the rendering loop already iterates all findings, the severity width can be computed the same way `locWidth` is, removing the manual constant and its unenforced invariant.

.mise/tasks/update-test-snapshots
```diff
@@ -0,0 +1,34 @@
...
+
+echo "Verifying the suite passes against the regenerated snapshots..."
+if ! go test -C "$root" ./...; then
+  echo
+  echo "Tests still fail after regeneration — this is a real failure, not a snapshot update (see output above)." >&2
+  exit 1
+fi
+
+echo
+echo "Snapshots regenerated — review the diff before committing."
```
iamrz1, line 34: The regeneration relies on two independent triggers — `UPDATE_SNAPSHOTS` for cupaloy and `CI`-unset for `pkg/stenciltest` — and the comment itself notes that running this locally may also refresh unrelated stenciltest template snapshots. Consider surfacing the workflow near the tests or in CONTRIBUTING, and tracking convergence of the two triggers, so the "review the diff before committing" step and the stenciltest side effect aren't discoverable only by reading this task file.

iamrz1 (changes requested):

.mise/tasks/update-test-snapshots
malept: Leaving the workflow documented in the task file for now. pkg/stenciltest's CI-unset trigger is on its way out with that package's replacement, so converging the two triggers isn't worth tracking; the cupaloy path stays on UPDATE_SNAPSHOTS.

internal/lint/manifest/manifest_test.go
malept: Replaced the constant with a dynamically computed severity width, so a longer severity value can no longer misalign the goldens.

internal/lint/manifest/manifest_test.go
malept: Added a comment noting the positional match relies on the guaranteed sorted finding order.

iamrz1 (approved):

.mise/tasks/update-test-snapshots
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+#MISE description="Regenerate all cupaloy test snapshots (UPDATE_SNAPSHOTS=true go test ./...)."
```
malept: The description equates this task with `UPDATE_SNAPSHOTS=true go test ./...`, but the script actually runs a regeneration pass followed by a separate verification pass (the block comment below explains why). Shortening the description avoids implying the task is a single command.

```suggestion
#MISE description="Regenerate all cupaloy test snapshots (regenerates goldens, then verifies)."
```

malept (commented):
A self-run of deepreview only found a docs nit.

---
🤖 Review generated by AI

.mise/tasks/update-test-snapshots
malept:
```suggestion
#MISE description="Regenerate and verify all cupaloy test snapshots"
```

PR #499, opened 2026-07-15
feat(lint): add `stencil lint templates --fix` and unify aggregate `--fix`

malept:

Adds `--fix` to `stencil lint templates`, mechanically migrating deprecated legacy block syntax (`###Block`/`###EndBlock`, in any of its three comment styles) to the v2 `<<Stencil::Block>>` form and re-linting so only genuinely unfixable findings remain. It also unifies the aggregate `stencil lint --fix`, which previously fixed the manifest only and silently skipped templates; it now fixes and re-lints both in one pass.

The fixer only touches syntax it's certain is safe: a dynamic-name legacy block (e.g. `###Block({{ $b }})`) is left alone because the linter can't even recognize it as a block today, and structural problems (missing `file.Block`, dangling or nested blocks) are never auto-fixed, only re-reported after the syntax rewrite.

## Jira ID

[DT-4828]

## Notes for your reviewers

`fixAndRelint` for the manifest subcommand now delegates to a new `fixManifestBytes` helper that returns findings instead of deciding pass/fail itself, so the aggregate path can combine manifest and template findings into one exit-code decision — worth a look if you're touching that code path again.

Verified manually against a built binary and a module using all three legacy comment styles plus a fixable manifest deprecation: stdin fix, in-place file fix, and the unified aggregate fix all produced correct output and were idempotent on a second run.

[DT-4828]: https://outreach-io.atlassian.net/browse/DT-4828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/stencil/lint.go
+++ b/cmd/stencil/lint.go
@@ -3,17 +3,11 @@
 // Description: This file contains the stencil lint command, which statically
 // validates a Stencil module without resolving dependencies.
 //
-// gosec G304 note: every os.Open/os.ReadFile/os.WriteFile call in this file
-// takes a path that is either a CLI argument the invoking user typed directly,
-// or derived from one (e.g. joining a user-named directory with
-// "manifest.yaml"). Each path-accepting function runs it through
-// filepath.Clean before using it, resolving any "." or ".." elements
-// lexically. This is not a security boundary -- stencil lint's whole purpose
-// is to read/write files the local user already has filesystem access to,
-// the same trust model as `cat`/`grep`/`eslint <path>`, so there is no
-// privilege boundary here for a path to cross, and Clean does not (and is not
-// meant to) prevent the user from naming a path outside any particular
-// directory. It normalizes the path representation before use.
+// Every file path here is a CLI argument the user typed, or derived from one.
+// Each path-accepting function runs it through filepath.Clean before use, but
+// that's normalization, not a security boundary: there's no privilege
+// boundary for a path to cross in a local CLI tool operating on paths the
+// invoking user already has filesystem access to.
 
 package main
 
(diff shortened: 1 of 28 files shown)


cmd/stencil/lint.go
```diff
@@ -304,6 +375,42 @@ func runLintAggregate(_ context.Context, c *cli.Command) error {
...
 
+	if c.Bool("fix") {
+		var all []lint.Finding
+
+		fixPath, finding, err := resolveManifestPath(filepath.Join(dir, "manifest.yaml"))
+		if err != nil {
+			return errors.Wrap(err, "lint failed")
+		}
+		if finding == nil {
+			raw, readErr := os.ReadFile(fixPath) // path was cleaned inside resolveManifestPath.
```
malept, line 384: issue: This and one other location is triggering the following in Wiz: "Improper Path Validation in File Operations". However:

1. Per the comment, this was mitigated in `resolveManifestPath` by using `filepath.Clean`
2. This is for a CLI, with a user supplied path.

Additionally, Claude decided to put a file-level comment here with a bunch more details on this.

cmd/stencil/lint.go
```diff
@@ -95,20 +108,40 @@ func newLintTemplatesCommand() *cli.Command {
...
 			return errors.New("'-' expects piped input, not an interactive terminal")
 		}
+		if fix {
+			raw, err := io.ReadAll(c.Reader)
+			if err != nil {
+				return errors.Wrap(err, "failed to read stdin")
+			}
+			// Fixed template goes to stdout; diagnostics go to the logger
+			// (stderr), mirroring `lint module-manifest - --fix`.
+			fixed, applied := linttemplates.FixBytes("<stdin>", raw)
```
marnagy: Make a const with value `"<stdin>"` and pass it to the function for more clarity.

I noticed that this const is all over the file. Is there a reason why it is not a const?

cmd/stencil/lint.go
```diff
@@ -520,24 +629,40 @@ func failManifest(log logrus.FieldLogger, name string, findings []lint.Finding,
...
-		if err == nil && bytes.Equal(current, fixed) {
+		if bytes.Equal(original, fixed) {
 			return nil // no change: leave the file untouched
 		}
 		mode := os.FileMode(0o644)
 		if info, statErr := os.Stat(path); statErr == nil {
 			mode = info.Mode().Perm()
 		}
-		return os.WriteFile(path, fixed, mode)
+		return os.WriteFile(path, fixed, mode) // path was cleaned above.
```
marnagy:
```suggestion
		return os.WriteFile(path, fixed, mode)
```

Unnecessary comment, imo - `Clean()` is called in the beginning of 10 line function

cmd/stencil/lint.go
malept: > Is there a reason why it is not a const?

Because the linter didn't say to do so. Claude has updated it.

marnagy (approved):

PR #501, opened 2026-07-17
feat(lint): add suggestions to fix warnings/errors

malept:

`stencil lint` reported problems without saying what to do about them. Findings now end with a concrete next step: ones `--fix` already resolves (argument `type`/`values`, module `prerelease`, and — since #499 — legacy `###Block` template syntax) point to `--fix`; ones it can't (missing `name`, invalid `stencilVersion`, `required`+`default` conflicts, deprecated module `url`, empty/malformed manifests) get the manual fix spelled out instead. Module `url` is called out as not auto-fixable, since there's no mechanical way to derive an import-path `name` from an arbitrary URL.

Third-party message text (the JSON-schema compiler's errors) and the block-misuse messages shared verbatim with the codegen runtime are left untouched, to avoid coupling stencil's tests to a dependency's wording or diverging from runtime error output.

## Jira ID

[DT-4828]

[DT-4828]: https://outreach-io.atlassian.net/browse/DT-4828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/lint/manifest/manifest.go
+++ b/internal/lint/manifest/manifest.go
@@ -100,9 +100,11 @@ func Validate(res *LoadResult) []lint.Finding {
 	// Check 1: strict decode succeeded.
 	if strictErr != nil {
 		if errors.Is(strictErr, io.EOF) {
-			f.Errorf("manifest.yaml", "manifest is empty")
+			f.Errorf("manifest.yaml", "manifest is empty; add at least a 'name' field "+
+				"(e.g. 'name: github.com/getoutreach/stencil-base')")
 		} else {
-			f.Errorf("manifest.yaml", "invalid manifest: %v", strictErr)
+			f.Errorf("manifest.yaml", "invalid manifest: %v; check the YAML syntax and "+
+				"field types near this location", strictErr)
 		}
 	}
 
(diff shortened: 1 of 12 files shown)


internal/lint/manifest/manifest.go
```diff
@@ -100,9 +100,11 @@ func Validate(res *LoadResult) []lint.Finding {
 	// Check 1: strict decode succeeded.
 	if strictErr != nil {
 		if errors.Is(strictErr, io.EOF) {
-			f.Errorf("manifest.yaml", "manifest is empty")
+			f.Errorf("manifest.yaml", "manifest is empty; add at least a 'name' field "+
+				"(e.g. 'name: github.com/getoutreach/stencil-base')")
```
marnagy, line 104:
```suggestion
			f.Errorf("manifest.yaml", "manifest is empty; add at least a 'name' field "+
				"(e.g. 'name: github.com/getoutreach/stencil-base') and 'description' field")
```

I don't think we ever discussed what should a "minimal `service.yaml`" file look like. I think including description makes sense as well.

marnagy (approved):

internal/lint/manifest/manifest.go
malept: Manifests (i.e., `manifest.yaml`) don't have a description field currently. If we want to add one we should discuss it outside of this PR.

PR #503, opened 2026-07-21
feat(lint): add offline `stencil lint project-manifest`

malept:

Adds `stencil lint project-manifest`, an offline linter for a project's `service.yaml`, so common misconfigurations are reported up front with actionable messages instead of surfacing as a confusing failure later. It also runs as part of `stencil lint <dir>` when a `service.yaml` is present. The implementation deliberately mirrors the existing module-manifest linter, so finding output, suggestion wording, and line annotation stay consistent between the two.

One design choice worth understanding: the decode is lenient rather than strict. A `service.yaml`'s argument keys are defined by the modules it loads, so rejecting unknown keys would flag valid input. For the same reason, `arguments` and `replacements` are not validated in this PR — they can only be checked against resolved modules, which is online work in a follow-up. This PR is offline-only, with no `--fix` yet.

## Jira ID

## Notes for your reviewers

A missing file is treated differently by the subcommand and the aggregate, on purpose: `stencil lint project-manifest` errors when its target `service.yaml` is absent, while `stencil lint <dir>` skips a missing one, since a module may legitimately have none. This matches the existing module-manifest runner.

### DIFF
--- a/internal/lint/projectmanifest/projectmanifest.go
+++ b/internal/lint/projectmanifest/projectmanifest.go
@@ -14,7 +14,10 @@ import (
 	"bytes"
 	"errors"
 	"io"
+	"sort"
+	"strconv"
 
+	semver "github.com/Masterminds/semver/v3"
 	"go.yaml.in/yaml/v3"
 
 	"github.com/getoutreach/stencil/internal/lint"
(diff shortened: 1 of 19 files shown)


internal/lint/projectmanifest/projectmanifest.go
```diff
@@ -0,0 +1,186 @@
...
+	return "modules[" + strconv.Itoa(i) + "]"
+}
+
+// checkVersions implements F7: a versions entry with an empty-string value.
+func checkVersions(f *lint.Findings, mf *configuration.ServiceManifest) {
+	names := make([]string, 0, len(mf.Versions))
+	for name := range mf.Versions {
+		names = append(names, name)
+	}
+	sort.Strings(names)
```
marnagy:
```suggestion
	names := slices.Sorted(maps.Keys(mf.Versions))
```
Go 1.25 (this module) lets `checkVersions` collapse the make/append/`sort.Strings` block into one expression. This also drops the `sort` import in favor of `maps`/`slices`; same deterministic ordering. 
- Generated with AI

internal/lint/projectmanifest/resolve.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	if top == nil || top.Kind != yaml.MappingNode {
+		return 0
+	}
+
+	if path == "modules" || strings.HasPrefix(path, "modules.") || strings.HasPrefix(path, "modules[") {
+		return resolveModulePath(top, path)
+	}
+	// arguments.<key>, versions.<key>, replacements.<key>: opaque flat leaf key.
+	for _, p := range []string{"arguments.", "versions.", "replacements."} {
```
marnagy: Suggestion: this allocates a fresh `[]string{"arguments.", "versions.", "replacements."}` on every call to `resolvePath`. Hoist it to a package-level `var opaqueLeafPrefixes = []string{...}` so it is allocated once and reads as intent.
- Generated with AI

marnagy (approved):

PR #504, opened 2026-07-22
feat(lint): validate project-manifest arguments against resolved modules

malept:

Extends `stencil lint project-manifest` from offline-only checks to validating a `service.yaml`'s arguments against the modules it uses: schema conformance, required-but-unset arguments, and `from:` references across modules. Problems that previously appeared only as a render-time failure now surface at lint time, attributed to the module that imposed the constraint.

Resolution is online by default; a new `--offline` flag restricts the command to the syntactic checks from the previous PR, and piped stdin input is always treated as offline. When a module set fails to resolve, or a module's manifest can't be read, the command reports an error and skips the argument checks for that run instead of validating against an incomplete set of modules.

## Jira ID

## Notes for your reviewers

Stacked on the offline PR (#503), review that first. The online paths are exercised with local module replacements, so the suite never touches the network.

The argument validation reimplements the render-time logic rather than calling into it, because the render path is coupled to template execution and can fetch external schema references over the network, which a linter must not do. One further divergence from render, noted in the code: an explicit `null` argument value is treated as unset.

The aggregate `stencil lint <dir>` runs this phase alongside its existing runners rather than through the shared runner interface, since resolution needs a request context and the offline flag and that interface can't carry either.

### DIFF
--- a/cmd/stencil/lint.go
+++ b/cmd/stencil/lint.go
@@ -441,21 +441,34 @@ func runLintAggregate(ctx context.Context, c *cli.Command) error {
 	// the runner signature can't carry. A missing service.yaml is skipped so a
 	// module without one is "nothing to lint". Kept last to preserve cross-runner
 	// order: manifest, templates, then service.yaml.
-	sp := filepath.Join(dir, "service.yaml")
-	if _, statErr := os.Stat(sp); statErr == nil {
-		raw, readErr := os.ReadFile(sp)
-		if readErr != nil {
-			return errors.Wrapf(readErr, "failed to read %q", sp)
-		}
-		pf, err := runProjectManifestReaderOnline(ctx, log, sp, bytes.NewReader(raw), c.Bool("offline"))
-		if err != nil {
-			return errors.Wrap(err, "lint failed")
-		}
-		all = append(all, pf...)
+	pf, err := lintServiceYaml(ctx, log, filepath.Join(dir, "service.yaml"), c.Bool("offline"))
+	if err != nil {
+		return err
 	}
+	all = append(all, pf...)
 
 	logFindings(log, all)
 	return failIfFindings(all, c.Bool("warnings-as-errors"))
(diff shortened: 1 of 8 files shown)


cmd/stencil/lint.go
```diff
@@ -672,31 +720,62 @@ func runLintProjectManifest(_ context.Context, c *cli.Command) error {
...
-		if closer != nil {
-			defer closer.Close()
+			return append(lintprojectmanifest.Validate(res), lint.Finding{
+				Severity: lint.SeverityError,
+				Path:     "modules." + m.Name,
+				Message:  fmt.Sprintf("failed to read module %q manifest: %v", m.Name, err),
+			})
 		}
-		return runProjectManifestReader(log, path, r)
+		mfCopy := mf // per-iteration allocation: Manifest returns by value; avoid aliasing.
```
marnagy: **Redundant defensive copy — `&mf` is already safe here**

`mf` is declared with `:=` inside the loop body (line 763), so each iteration gets a fresh variable and `&mf` is a distinct pointer per iteration. The `mfCopy := mf` copy — and its "avoid aliasing" comment — guards against a hazard that doesn't apply here: the aliasing footgun is specific to the `range` variable `m` itself, and even that was fixed in Go 1.22. Body-scoped `:=` variables have always gotten a fresh address per iteration when their address escapes.

This can be simplified to:

```go
resolved = append(resolved, lintprojectmanifest.ResolvedModule{
    ImportPath: m.Name,
    Module:     m,
    Manifest:   &mf,
})
```

The same premise underlies the `ResolvedModule` doc comment in `internal/lint/projectmanifest/projectmanifest.go` ("the command layer MUST allocate one manifest per module … or every pointer would alias the last one") — that's incorrect for a body-scoped `:=` variable, so the note can be trimmed too.

- Generated with AI

marnagy (approved):

PR #505, opened 2026-07-22
feat(lint): flag replacement and cross-module argument problems

malept:

Adds four online checks to `stencil lint project-manifest`. Two cover `replacements`: a key that matches no module in the dependency graph is reported as inert, and a key that does match a module but points at a local path that doesn't exist is reported as an error. The other two cover arguments across modules: an argument provided in `service.yaml` that no resolved module declares is flagged as a likely typo, and two modules declaring the same argument with conflicting schemas are flagged so the mismatch can be reconciled.

All four are warnings except the missing-replacement-path case, which is an error. The undeclared-argument check is suppressed when a module declares a catch-all argument, since such a module can legitimately consume keys it doesn't name.

## Jira ID

## Notes for your reviewers

Stacked on the online-core PR ([#504](https://github.com/getoutreach/stencil/pull/504)); review that first. The checks are additive: they append to the existing online findings and leave the previous PR's resolution and argument-value logic unchanged.

The undeclared-argument check matches at the top level. A dotted declared name such as `aws.IRSA`, provided as a nested `aws:` object, is not flagged; a mistyped nested key under a valid top-level key is left to schema validation rather than reported here. That per-leaf case is noted in the code as future work.

The missing-local-replacement-path check has no command-level test because the scenario can't be reached end to end: a matched module whose replacement path is missing fails resolution before this check runs. It is covered by package tests that supply resolved modules directly.

### DIFF
--- a/internal/lint/projectmanifest/online.go
+++ b/internal/lint/projectmanifest/online.go
@@ -9,9 +9,12 @@ package projectmanifest
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"maps"
 	"os"
+	"slices"
 	"sort"
 	"strings"
 
@@ -79,13 +82,7 @@ func checkReplacements(res *LoadResult, mods []ResolvedModule) []lint.Finding {
 		resolved[m.ImportPath] = struct{}{}
 	}
 
-	keys := make([]string, 0, len(res.Manifest.Replacements))
-	for k := range res.Manifest.Replacements {
-		keys = append(keys, k)
-	}
-	sort.Strings(keys)
-
-	for _, key := range keys {
+	for _, key := range slices.Sorted(maps.Keys(res.Manifest.Replacements)) {
 		value := res.Manifest.Replacements[key]
 		if _, matched := resolved[key]; !matched {
 			// O5: inert replacement key.
@@ -97,7 +94,7 @@ func checkReplacements(res *LoadResult, mods []ResolvedModule) []lint.Finding {
 		// Matched key: O8 checks a local path's existence.
 		if uriIsLocal(value) {
 			path := strings.TrimPrefix(value, "file://")
-			if _, err := os.Stat(path); err != nil {
+			if _, err := os.Stat(path); errors.Is(err, os.ErrNotExist) {
 				f.Errorf("replacements."+key,
 					"replacement for %q points at a local path that does not exist: %s; "+
 						"create the path, or correct the replacement", key, path)
@@ -137,13 +134,7 @@ func checkUndeclaredArgs(res *LoadResult, idx map[string][]declaration, mods []R
 		declaredTop[top] = struct{}{}
 	}
 
-	keys := make([]string, 0, len(res.Manifest.Arguments))
-	for k := range res.Manifest.Arguments {
-		keys = append(keys, k)
-	}
-	sort.Strings(keys)
-
-	for _, key := range keys {
+	for _, key := range slices.Sorted(maps.Keys(res.Manifest.Arguments)) {
 		if _, ok := declaredTop[key]; ok {
 			continue // this top-level key is (a prefix of) a declared name
 		}
@@ -170,16 +161,17 @@ func hasDynamicArg(mods []ResolvedModule) bool {
 			if t != "object" {
 				continue
 			}
-			ap, present := arg.Schema["additionalProperties"]
-			// Open object: additionalProperties absent (default true) or explicitly true / a schema.
-			if !present {
-				return true
-			}
-			if b, ok := ap.(bool); ok && b {
-				return true
-			}
-			if _, isSchema := ap.(map[string]interface{}); isSchema {
-				return true
+			// Open object: additionalProperties absent (default true), explicitly
+			// true, a sub-schema, or present-but-null (treated as open).
+			switch ap := arg.Schema["additionalProperties"].(type) {
+			case nil:
+				return true // key absent, or present-but-null → open
+			case bool:
+				if ap {
+					return true // additionalProperties: true → open
+				}
+			case map[string]interface{}:
+				return true // additionalProperties is a sub-schema → open
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


internal/lint/projectmanifest/online.go
```diff
@@ -54,6 +56,136 @@ func sortOnline(findings []lint.Finding) {
...
+			// O5: inert replacement key.
+			f.Warnf("replacements."+key,
+				"replacement for %q matches no module in the dependency graph and is "+
+					"ignored; remove it, or correct the import path to a resolved module", key)
+			continue
+		}
+		// Matched key: O8 checks a local path's existence.
+		if uriIsLocal(value) {
+			path := strings.TrimPrefix(value, "file://")
+			if _, err := os.Stat(path); err != nil {
```
marnagy: **Fix (point 1): narrow the `os.Stat` error to genuine absence.** `os.Stat` returns a non-nil error for permission (`EACCES`), symlink-loop, and I/O failures too, not only absence. Each of those currently emits "points at a local path that does not exist", sending the user to fix the wrong thing. Per Effective Go's error guidance (https://go.dev/doc/effective_go#errors), inspect the error instead of collapsing it to a boolean:

```suggestion
			if _, err := os.Stat(path); errors.Is(err, os.ErrNotExist) {
```

This needs `"errors"` added to the import block. On a permission/IO error the check now stays silent, which is the safer default for a linter than a false "does not exist".

- Generated with AI

internal/lint/projectmanifest/online.go
```diff
@@ -54,6 +56,136 @@ func sortOnline(findings []lint.Finding) {
...
+		resolved[m.ImportPath] = struct{}{}
+	}
+
+	keys := make([]string, 0, len(res.Manifest.Replacements))
+	for k := range res.Manifest.Replacements {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	for _, key := range keys {
```
marnagy: **Fix (point 4): fold the collect-sort-range boilerplate into one line.** This "make slice -> range map -> append -> sort -> range" shape repeats verbatim in `checkReplacements`, `checkUndeclaredArgs` (~L140), and `checkSchemaConflicts` (~L289). On Go 1.25 the stdlib expresses the intent directly, in the spirit of Effective Go's preference for clear control flow (https://go.dev/doc/effective_go#control-structures):

```suggestion
	for _, key := range slices.Sorted(maps.Keys(res.Manifest.Replacements)) {
```

Add `"maps"` and `"slices"` to the imports, and apply the same collapse at the other two sites (they range over `res.Manifest.Arguments` and `idx`). No behavior change: `slices.Sorted` orders strings identically to `sort.Strings`.

- Generated with AI

internal/lint/projectmanifest/online.go
```diff
@@ -54,6 +56,136 @@ func sortOnline(findings []lint.Finding) {
...
+			// Open object: additionalProperties absent (default true) or explicitly true / a schema.
+			if !present {
+				return true
+			}
+			if b, ok := ap.(bool); ok && b {
+				return true
+			}
+			if _, isSchema := ap.(map[string]interface{}); isSchema {
+				return true
+			}
```
marnagy: **Fix (point 5): replace the type-discriminating `if` cascade with a type switch.** This is the exact shape Effective Go's type switch is for (https://go.dev/doc/effective_go#type_switch), and it also resolves a latent ambiguity: today `additionalProperties: null` (key present, value nil) falls through all three `if`s and is treated as *closed*, identical to `additionalProperties: false`, silently. The `case nil` arm below makes both absent and null mean "open" as a deliberate, testable choice rather than an accident of control flow:

```suggestion
			switch ap := arg.Schema["additionalProperties"].(type) {
			case nil:
				return true // key absent, or present-but-null -> treat as open
			case bool:
				if ap {
					return true // additionalProperties: true -> open
				}
			case map[string]interface{}:
				return true // additionalProperties is a sub-schema -> open
			}
```

Worth adding a test for the `additionalProperties: null` input to pin the intent.

- Generated with AI

internal/lint/projectmanifest/online.go
```diff
@@ -148,6 +280,71 @@ func resolveFrom(f *lint.Findings, mods []ResolvedModule, owner ResolvedModule,
...
+	bb, err := marshalSorted(b)
+	if err != nil {
+		return false
+	}
+	return bytes.Equal(ab, bb)
+}
+
+// marshalSorted JSON-marshals v with map keys sorted (encoding/json already
+// sorts map keys), returning canonical bytes for comparison.
+func marshalSorted(v map[string]interface{}) ([]byte, error) {
```
marnagy: Suggestion: (point 3) `marshalSorted` is a no-op wrapper. It only calls `json.Marshal(v)`, and its own comment admits "encoding/json already sorts map keys". The name implies a sort step the body does not perform, which Effective Go's naming guidance (https://go.dev/doc/effective_go#names) argues against. Consider inlining `json.Marshal(a)` / `json.Marshal(b)` at the two call sites in `schemaEquivalent` and deleting this function, then move the guarantee actually relied on (encoding/json sorts map keys recursively and preserves slice order, so the byte-compare is canonical) into `schemaEquivalent`'s doc, where the correctness lives.

- Generated with AI

marnagy (commented):
Automated review: 3 committable fixes (points 1, 4, 5) and 1 suggestion (point 3), grounded in Effective Go (https://go.dev/doc/effective_go). Each inline comment stands on its own.

- Generated with AI

internal/lint/projectmanifest/online.go
malept: Narrowed the check to `errors.Is(err, os.ErrNotExist)` and added `"errors"` to the imports, so O8 now fires only on genuine absence and stays silent on permission/IO errors. No test was added for the non-absence case, since reliably provoking an `EACCES`/IO stat error isn't portable across CI environments; the absence path remains covered by `TestCheckReplacementsMatchedLocalMissingO8`.

----
:robot: _Generated with AI_

marnagy (approved):

PR #506, opened 2026-07-22
feat(lint): add `--fix` to `stencil lint project-manifest`

malept:

`stencil lint project-manifest` already warns when a `service.yaml` module uses the deprecated `prerelease` field and points users at `--fix`, but the subcommand didn't yet accept that flag. This wires it up: `--fix` migrates `prerelease: true` to `channel: rc` (preserving an existing `channel`) and drops the redundant `prerelease: false`, editing the YAML node tree in place so comments and key order survive. After fixing, it re-lints in the ambient mode — online by default, offline under `--offline`, and always offline for stdin (`-`), which writes the fixed document to stdout. The aggregate `stencil lint --fix` gains the same `service.yaml` migration.

Because a project manifest and a template repository manifest share the same module shape, the module fixer is extracted into a new shared `internal/lint/modulefix` package that both linters consume; the manifest fixer's behavior is unchanged (its snapshots are byte-identical).

This is the final change in the `project-manifest` lint series and builds on #505.

## Jira ID

[DT-5396]

## Notes for your reviewers

The `Applied` types in the manifest and project-manifest packages are now aliases of `modulefix.Applied`, so the existing fix-logging helper accepts both without a signature change. A no-op fix returns the original bytes verbatim and skips the write, so an already-clean file keeps its mtime.

### DIFF
--- a/cmd/stencil/lint.go
+++ b/cmd/stencil/lint.go
@@ -356,6 +356,52 @@ func failTemplates(log logrus.FieldLogger, findings []lint.Finding, warningsAsEr
 	return failLint(log, "templates are", findings, warningsAsErrors)
 }
 
+// runLintAggregateFix fixes a module's manifest.yaml, template files, and
+// service.yaml in place (each skipped when absent), then logs and applies the
+// warnings-as-errors policy to whatever findings survive. It mirrors the
+// non-fix aggregate runner order: manifest, templates, then service.yaml.
+func runLintAggregateFix(ctx context.Context, c *cli.Command, log logrus.FieldLogger, dir string) error {
+	var all []lint.Finding
+
+	fixPath, finding, err := resolveManifestPath(filepath.Join(dir, "manifest.yaml"))
+	if err != nil {
+		return errors.Wrap(err, "lint failed")
+	}
+	if finding == nil {
+		raw, readErr := os.ReadFile(fixPath) // path was cleaned inside resolveManifestPath.
+		if readErr != nil {
+			return errors.Wrapf(readErr, "failed to read %q", fixPath)
+		}
+		findings, fixErr := fixManifestBytes(log, fixPath, raw, writeFixedFile(fixPath, raw))
+		if fixErr != nil {
+			return errors.Wrap(fixErr, "lint failed")
+		}
+		all = append(all, findings...)
+	}
+	// A missing manifest is skipped, not fixed: aggregate lint treats it as
+	// "nothing to lint" (a module may be templates-only), matching the
+	// non-fix aggregate path and manifestRunner.
+
+	files, err := collectTemplateFiles(log, filepath.Join(dir, templatesDir))
+	if err != nil {
+		return errors.Wrap(err, "lint failed")
+	}
+	findings, err := fixTemplateFiles(log, files)
+	if err != nil {
+		return errors.Wrap(err, "lint failed")
+	}
+	all = append(all, findings...)
+
+	pf, err := fixServiceYaml(ctx, log, filepath.Join(dir, "service.yaml"), c.Bool("offline"))
+	if err != nil {
+		return err
+	}
+	all = append(all, pf...)
+
+	logFindings(log, all)
+	return failLint(log, fmt.Sprintf("module %q is", dir), all, c.Bool("warnings-as-errors"))
+}
+
 // runLintAggregate is the `stencil lint [dir]` action.
 func runLintAggregate(ctx context.Context, c *cli.Command) error {
 	if c.Args().Len() > 1 {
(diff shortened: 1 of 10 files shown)


internal/lint/manifest/manifest.go
```diff
@@ -226,8 +226,8 @@ func checkModules(f *lint.Findings, mf *configuration.TemplateRepositoryManifest
 }
 
 // moduleIDPath builds the finding path for module i, preferring its name over
-// its slice index. Shared by the checker (modulePath) and the fixer
-// (moduleFixPath) so their finding paths stay identical.
+// its slice index. Kept identical to modulefix.ModulePath so the checker's and
+// the fixer's finding paths stay in sync.
```
marnagy: **Fix: delegate to `modulefix.ModulePath` instead of hand-syncing this formatter.**

Now that `modulefix` owns `ModulePath`, this hand-maintained copy is exactly the drift risk the shared package was created to remove — the fixer already calls `modulefix.ModulePath`; only the checker keeps its own copy. Replace the body with a delegation:

```go
// moduleIDPath builds the finding path for module i, preferring its name over
// its slice index. Delegates to modulefix.ModulePath so the checker and the
// fixer produce identical paths from a single implementation.
func moduleIDPath(name string, i int) string {
	return modulefix.ModulePath(name, i)
}
```

Two accompanying edits in this file's import block:
- add `"github.com/getoutreach/stencil/internal/lint/modulefix"` to the getoutreach import group
- remove `"strconv"` — it becomes unused

(Applied locally and confirmed `go build ./...` and `go test ./internal/lint/...` pass.)

- Generated with AI

cmd/stencil/lint.go
```diff
@@ -778,6 +841,60 @@ func resolveAndValidateProjectManifest(ctx context.Context, log logrus.FieldLogg
...
+func fixProjectManifestBytes(ctx context.Context, log logrus.FieldLogger, name string,
+	raw []byte, offline bool, writeFixed func([]byte) error) ([]lint.Finding, error) {
+	fixed, applied, ok := lintprojectmanifest.FixBytes(raw)
+	if !ok {
+		return runProjectManifestReaderOnline(ctx, log, name, bytes.NewReader(raw), offline)
+	}
+	if err := writeFixed(fixed); err != nil {
+		return nil, errors.Wrap(err, "failed to write fixed service manifest")
+	}
+	logApplied(log, applied)
```
marnagy, line 906: Suggestion: type this shared helper's parameter as the shared type.

`logApplied` (defined at `cmd/stencil/lint.go:949` as `[]lintmanifest.Applied`) accepts the project-manifest `applied` here only because both packages' `Applied` are aliases of `modulefix.Applied`. That compiles, but the signature reads as manifest-specific. Retyping the parameter `[]modulefix.Applied` expresses the real contract — it logs any module fixer's output — and stops it looking coupled to the manifest linter.

- Generated with AI

cmd/stencil/lint.go
```diff
@@ -471,6 +485,26 @@ func lintServiceYaml(ctx context.Context, log logrus.FieldLogger,
 	return pf, nil
 }
 
+// fixServiceYaml is the aggregate --fix counterpart to lintServiceYaml: it
+// migrates a module's service.yaml in place (if present) and returns the
+// post-fix findings. A missing service.yaml is skipped (nil, nil), matching
+// lintServiceYaml and the non-fix aggregate path.
+func fixServiceYaml(ctx context.Context, log logrus.FieldLogger,
```
marnagy, line 511: Suggestion: `fixServiceYaml` is a near-exact clone of `lintServiceYaml` (just above) — identical stat/skip/read prologue, differing only in the single processing call. Consider extracting the shared prologue so both paths share one implementation:

```go
// readServiceYamlOrSkip returns the file bytes, or skip=true when service.yaml is absent.
func readServiceYamlOrSkip(sp string) (raw []byte, skip bool, err error) {
	if _, statErr := os.Stat(sp); statErr != nil {
		return nil, true, nil
	}
	raw, err = os.ReadFile(sp)
	return raw, false, errors.Wrapf(err, "failed to read %q", sp)
}
```

Judgment call — the package deliberately mirrors the manifest linter with parallel `fix*`/`lint*` pairs, so I'd only de-duplicate this pure boilerplate, not the whole family.

- Generated with AI

cmd/stencil/lint.go
```diff
@@ -471,6 +485,26 @@ func lintServiceYaml(ctx context.Context, log logrus.FieldLogger,
...
+	sp string, offline bool) ([]lint.Finding, error) {
+	if _, statErr := os.Stat(sp); statErr != nil {
```
marnagy: Suggestion: this skips on any `os.Stat` error, so a permission error (or other non-`NotExist` failure) on service.yaml is silently swallowed rather than surfaced. It matches the pre-existing `lintServiceYaml` (so not a regression), but `resolveProjectManifestPath` in this same file correctly distinguishes `os.IsNotExist` from other stat errors. If you touch this, gating the skip on `os.IsNotExist(statErr)` and returning the wrapped error otherwise would make the skip intentional.

- Generated with AI

marnagy (commented):
Automated follow-up review of the `project-manifest --fix` branch, from an analysis of the diff against the PR base. The branch matches its description; I found no correctness bugs. Below are two fixes (points 1 and 5) and three optional cleanups (points 2–4).

- Generated with AI

marnagy (approved):

PR #519, opened 2026-08-11
feat(configuration): add `refines:` field to `Argument`

malept:

`stencil lint project-manifest` warns when two modules declare the same argument with different schemas. This is a false positive when a dependent module deliberately narrows a foundational module's schema. This stack adds a `refines:` field to state that intent, and a lint check to verify it. This PR adds the field itself and its single-manifest validation. The schema-narrowing check and the check's online wiring follow in the next two PRs.

`refines:` names a dependency module that declares an argument of the same name. Unlike `from:`, the local schema is kept. `refines:` and `from:` are mutually exclusive. The offline `stencil lint` manifest check enforces this and rejects the combination. A `refines:` argument still has its local schema validated, unlike a `from:` argument, which defers entirely to the referenced module.

This PR alone has no observable lint effect: nothing yet checks that a `refines:` schema is a true narrowing. That check lands in the next PR.

## Jira ID

[DT-5444]

## Notes for your reviewers

- **Public struct addition (Ask boundary).** This adds one field, `Refines`, to the exported `configuration.Argument` type. The addition is backward compatible: a new optional YAML key. It is also registered in the manifest fixer's `knownArgFields`, so `--fix` does not sweep a `refines:` key into `schema`.

[DT-5444]: https://outreach-io.atlassian.net/browse/DT-5444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/lint/manifest/manifest_test.go
```diff
@@ -323,6 +323,52 @@ func TestValidateWholeDocumentFindingHasNoLine(t *testing.T) {
...
+name: github.com/getoutreach/stencil-golang
+modules:
+  - name: github.com/getoutreach/stencil-base
+arguments:
+  commands:
+    refines: github.com/getoutreach/stencil-base
+    schema:
+      type: 123
+`
+	res, err := lintmanifest.Load(strings.NewReader(y))
```
marnagy: Suggestion: This re-implements `Load` + `Validate` + a `NilError` check on the read error, but the file already has a `validateString(in string) []lint.Finding` helper used by every other `Validate`-focused test here (e.g. `TestValidateFromCarveOut`). Worth switching to `findings := validateString(y)` here to stay consistent with the file's existing convention and drop the redundant error check (`Load` only errors when reading from `r` fails, which a `strings.Reader` never does).

- Generated using AI

internal/lint/manifest/manifest_test.go
```diff
@@ -323,6 +323,52 @@ func TestValidateWholeDocumentFindingHasNoLine(t *testing.T) {
...
+	const y = `
+name: github.com/getoutreach/stencil-golang
+modules:
+  - name: github.com/getoutreach/stencil-base
+arguments:
+  commands:
+    from: github.com/getoutreach/stencil-base
+    refines: github.com/getoutreach/stencil-base
+`
+	res, err := lintmanifest.Load(strings.NewReader(y))
```
marnagy: Suggestion: Same as the comment above on `TestCheckArgumentsRefinesSchemaStillValidated` — use `findings := validateString(y)` here too instead of re-implementing `Load` + `Validate` + error check.

- Generated using AI

pkg/configuration/configuration.go
```diff
@@ -206,6 +206,15 @@ type Argument struct {
...
 	From string `yaml:"from"`
+
+	// Refines names another module (which this module MUST list in `modules:`)
+	// that declares an argument of the same name. Unlike `from:`, the local
+	// schema is KEPT; `refines:` asserts that the local schema is a valid
+	// narrowing (subschema) of the referenced module's schema for the same
+	// argument. The project-manifest linter verifies this and fails (error
+	// severity) if it does not hold. `refines:` and `from:` are mutually
+	// exclusive.
+	Refines string `yaml:"refines,omitempty"`
```
marnagy: Nitpick: `Refines` is tagged `yaml:"refines,omitempty"` while its mutually-exclusive sibling `From` (line 208) is tagged `yaml:"from"` with no `omitempty`. Harmless today since nothing round-trips `Argument` through `yaml.Marshal`, but worth aligning the tagging between these two related fields for consistency.

- Generated using AI

marnagy (changes requested):
Implementation matches the PR description; build and tests pass.
A few minor points inline (Suggestion/Nitpick), nothing blocking.

Nitpick: `docs/content/en/reference/template-module.md` documents `from:` but not `refines:` yet. Not part of this diff, so not left as a line comment — likely intentional given this is step 1 of a 3-PR stack, flagging as FYI.

- Generated using AI

marnagy (approved):

PR #530, opened 2026-09-22
feat(lint): flag & fix single "#" prefixes on Stencil::Block tags

malept:

`stencil lint templates` now catches a single `#` comment marker used in place of the required `##` before a `<<Stencil::Block(...)>>` tag (and its matching close tag), and `--fix` migrates it automatically. A single-`#` tag was previously invisible to every block-recognition regex in the linter, so it wasn't just unflagged — the block was treated as entirely absent, which cascaded into unrelated "never closed" and "illegal nesting" errors on every block that followed in the file. This surfaced from a real occurrence in a downstream service template (and, per review, is live in this repo's own `AGENTS.md`, whose `customCommands` block uses a single `#`).

## Jira ID

[DT-5572]

## Notes for your reviewers

The render pipeline's own regex (`codegen.V2BlockPattern`) is untouched; this is a lint-only addition, folded into the existing `v2StartAny`/`v2EndAny` supplemental regexes rather than added as separate patterns, so detection stays anchored to the tag's actual comment-prefix position and won't false-positive on a stray `#` elsewhere in a line. The fix is always safe to apply automatically (it only duplicates one character; it never has to guess which block a tag refers to, unlike the existing legacy-syntax migration's name-mismatch handling).

`dynamic` (whether a name is a template expression vs. a literal) is derived from the name text itself against `literalBlockName`, not from which regex branch matched — that branch is now reachable for a bad prefix too, not just a dynamic name. This is what lets rule 1 still offer its `Add {{ file.Block "name" }}` suggestion for a single-`#` block with a literal name, instead of the shorter dynamic-name wording.

Out of scope for this PR (no worse than `main`, tracked as follow-ups): a single-`#` tag spelled `EndBlock` instead of `Block` (e.g. `# <<Stencil::EndBlock>>`), and a triple-`#` (or other wrong-length) prefix. Neither matches any existing pattern today, on `main` or on this branch, so this PR doesn't regress them — it just doesn't close that broader class of "wrong prefix" defects yet.

Verified against a real downstream template in addition to the new unit tests.

---
_Generated by [Claude Code](https://claude.ai/code/session_017FN9LymAqJ5DVX355PzYTt)_

[DT-5572]: ``https://outreach-io.atlassian.net/browse/DT-5572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ``


malept:
Two checks are red on this PR's head (`acb6d72`): **Conventional Commit / run** and **deploy** (`github pages`). Neither is caused by this PR's diff — both are failing repo-wide right now:

- **deploy** (`github pages`) fails identically on direct pushes to `main` itself (e.g. runs for #526 and #527, both merged commits on `main`), completing in ~4s with no job logs available at all. The workflow only builds/deploys docs via Hugo; this PR doesn't touch `docs/`.
- **Conventional Commit / run** (a reusable workflow from `getoutreach/actions@main`) is failing on every recent PR I can see, including plain `chore(deps): bump ...` dependabot PRs (#528, #529) with unambiguously valid conventional-commit titles, also completing in ~2-4s with no downloadable logs.

Both fail too fast and too uniformly (no logs at all, not even empty ones) to be a problem in this PR's code — this looks like a repo/org-level CI infra issue (e.g. a broken shared-workflow reference or runner/token problem), not something a change to `internal/lint/templates` could cause or fix. I don't have a fix to port for either. I'll re-run both once to rule out simple transience; if they fail again I'll leave this PR as-is rather than working around them.

https://claude.ai/code/session_017FN9LymAqJ5DVX355PzYTt

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/lint/templates/templates.go
```diff
@@ -31,22 +31,37 @@ var fileBlockCall = regexp.MustCompile(`\{\{-?\s*.*\bfile\.Block\b`)
...
+// of the required "##" (rule 6). It is a lint-only supplement so a
+// dynamic-name and/or bad-prefix block still balances against its end tag
+// (avoiding a false "bare end tag"/"never closed") and still gets the
+// presence-based file.Block check. It deliberately does NOT match end tags
+// or EndBlock. Group 1 is the matched prefix (classify() checks it against
+// "#" to set singleHash); group 2 is the name/args, display-only. Anchoring
+// to the start of the line (mirroring V2BlockPattern) is what makes this
+// safe: a "#" elsewhere on the line can never match, only one sitting where
+// the line's own comment prefix belongs.
+var v2StartAny = regexp.MustCompile(`^\s*(//|##|--|<!--|#)\s?<<Stencil::Block(\([^>]*\))?>>`)
```
marnagy: Improvement: rule 6 covers only the `Block` spelling with exactly one `#`. Two neighbours in the same defect class stay invisible:

- `# <<Stencil::EndBlock>>` and `# <</Stencil::EndBlock>>` — neither regex can match, because both require the literal `Block` directly after `Stencil::`. `V2BlockPattern` misses them too (wrong prefix), so there is no rule 5 either. The tag becomes a plain comment at render time: the same silent data loss this PR closes.
- `### <<Stencil::Block(x)>>` (one `#` too many) — `V2BlockPattern` fails (`##`, then `#`, no `<<`), `BlockPattern` fails (`[a-zA-Z ]+` cannot match ` <<`), and both new regexes fail. This is a plausible half-finished legacy-to-v2 migration.

This is no worse than `main`. Either handle the two cases here, or record them as out of scope in the PR body.

- Generated using AI

internal/lint/templates/templates.go
```diff
@@ -309,17 +354,28 @@ func classify(text string) token {
...
-		// balances against its open instead of dangling. Mirrors v2StartAny.
-		return token{end: true}
+	if m := v2EndAny.FindStringSubmatch(text); m != nil {
+		// Mirrors v2StartAny: reached for a dynamic-name close, a bad
+		// single-"#" prefix (rule 6), or both. Recognizing it as a real end
+		// either way means a dynamic-name and/or bad-prefix block's close
+		// still balances against its open, instead of leaving it dangling
+		// and cascading into false "illegal nesting" errors on every block
+		// that follows.
+		singleHash := m[1] == "#"
```
marnagy: Fix: `# <</Stencil::Block(name)>>` loses its rule-5 finding.

With `##`, a closing tag that carries arguments reaches `codegen.V2BlockPattern` and reports `MsgClosingTagArgs`. With `#`, `V2BlockPattern` misses and `v2EndAny` accepts the line as a plain end tag, so only rule 6 fires. A plain `stencil lint templates` run therefore under-reports the misuse. `--fix` plus the re-lint does surface it, so only the no-fix path is affected.

Note that setting `misuse` in this branch is not a drop-in fix: `scan`'s `case tok.misuse != ""` arm is exclusive and would suppress rule 6 on the same line.

- Generated using AI

internal/lint/templates/templates.go
```diff
@@ -31,22 +31,37 @@ var fileBlockCall = regexp.MustCompile(`\{\{-?\s*.*\bfile\.Block\b`)
...
+// bad-prefix cases. Group 1 is the matched prefix.
+var v2EndAny = regexp.MustCompile(`^\s*(//|##|--|<!--|#)\s?<</Stencil::Block(\([^>]*\))?>>`)
+
+// literalBlockName matches a name satisfying codegen.V2BlockPattern's own
+// name character class -- i.e. a literal identifier, as opposed to a dynamic
+// template expression like "{{ $x }}" that the strict pattern rejects (and
+// that reaches v2StartAny/v2EndAny instead). Used to decide whether a name
+// reaching those lint-only regexes is dynamic, so that check stays tied to
+// the actual character class rather than a "{{" substring guess.
+var literalBlockName = regexp.MustCompile(`^[a-zA-Z0-9 _]+$`)
```
marnagy: Improvement: `^[a-zA-Z0-9 _]+$` hand-copies the argument character class inside `codegen.V2BlockPattern` (`internal/codegen/blocks.go:46`). `dynamic` stays correct only while the two classes are identical, and nothing enforces that. Export the class from `codegen` and build both patterns from it, or add a test that asserts the two agree.

- Generated using AI

internal/lint/templates/fix.go
```diff
@@ -109,6 +111,35 @@ func fixLine(line string, hasOpen bool, openName string) (fixed, message string)
...
+// block, which v2StartAny/v2EndAny also match).
+//
+// It reuses the same v2StartAny/v2EndAny regexes and prefix group (index
+// [2:4] of FindStringSubmatchIndex) that classify() uses for the linter's
+// own singleHash check, so this fixer can never drift from what rule 6
+// flags: fixing every single-"#" tag is always safe here because it's
+// mechanical (duplicate one character) and never guesses at intent, unlike
+// the legacy migration's name-mismatch caution above.
+func fixSingleHashLine(line string) (fixed, message string) {
+	if idx := v2StartAny.FindStringSubmatchIndex(line); idx != nil {
```
marnagy: Suggestion: use named capture groups instead of the positional `idx[2]:idx[3]`. This PR is the evidence for the risk: adding the prefix group already moved the name from `m[1]` to `m[2]` in `classify`.

```suggestion
	if idx := v2StartAny.FindStringSubmatchIndex(line); idx != nil {
		p := 2 * v2StartAny.SubexpIndex("prefix")
```

This needs the regexes in `templates.go` to name their groups, for example:

```go
var v2StartAny = regexp.MustCompile(`^\s*(?P<prefix>//|##|--|<!--|#)\s?<<Stencil::Block(?P<args>\([^>]*\))?>>`)
```

- Generated using AI

marnagy (changes requested):
Implementation matches the description. `codegen.V2BlockPattern` is untouched; detection is folded into `v2StartAny`/`v2EndAny`; the `^\s*` anchor makes a mid-line `#` unmatchable (covered by two negative tests); the fix is one inserted byte and is idempotent; `##` beats `#` in the alternation because Go's `regexp` is leftmost-first; `literalBlockName` mirrors `V2BlockPattern`'s arg class, so deriving `dynamic` is behaviour-preserving and no pre-existing snapshot changed. CircleCI is green.

Praise: this bug is live in this repo. `AGENTS.md:29-31` uses `# <<Stencil::Block(customCommands)>>`, so that block's content is discarded on every render.

Documentation: the PR body omits the `dynamic` derivation change (from "reached the `v2StartAny` branch" to "the name is not a literal"). That change is what gives rule 1 the actionable ``Add {{ file.Block "bad" }}`` suffix for a single-`#` literal-name block. One line in "Notes for your reviewers" would help the next reader.

- Generated using AI

internal/lint/templates/templates.go
malept: Good catch, and thanks for the note on the `misuse` exclusivity trap — that would have been a real regression. Fixed in f2551d3 with a separate `argsOnClose` token field checked independently inside `scan()`'s `tok.end` case (alongside `singleHash`, not through the exclusive `tok.misuse` arm), so both findings fire together. A close carrying a dynamic name still stays clean, mirroring the correctly-prefixed case. New tests: "single hash close tag with literal args also reports closing-tag-args misuse" and "...with dynamic args is not misuse".

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/lint/templates/templates.go
malept: Took the test option rather than exporting the class from `codegen` (which would be a public API change). Added `TestLiteralBlockNameMatchesV2BlockPatternArgClass` in a new `internal_test.go` (white-box, package `templates`): it probes every printable ASCII character as a single-char name and asserts `literalBlockName` agrees with what `codegen.V2BlockPattern`'s own args group actually accepts, so future drift between the two fails a test instead of silently changing `dynamic`'s behavior. Pushed in f2551d3.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/lint/templates/fix.go
malept: Agreed, and this PR is indeed the evidence for it. Named both groups (`prefix`, `args`) on `v2StartAny`/`v2EndAny` and switched every access to `SubexpIndex` — in `classify()` (templates.go) and `fixSingleHashLine` (fix.go), using your suggested `2 * v2StartAny.SubexpIndex("prefix")` pattern for the `FindStringSubmatchIndex` case. Pushed in f2551d3.

---
_Generated by [Claude Code](https://claude.ai/code)_

internal/lint/templates/templates.go
malept: Confirmed both are unreachable today, on `main` and on this branch (matches your analysis exactly — same leftmost-first regex-alternation failure mode for `###`, same missing-`Block`-literal gap for `EndBlock`-spelled tags). Recording both as explicitly out of scope in the PR body rather than folding them into this PR, since they're a distinct defect shape (wrong word / wrong prefix length, not "single # instead of double #") worth their own focused change. Not a regression either way.

---
_Generated by [Claude Code](https://claude.ai/code)_

malept:
@marnagy thanks for the thorough review. Summary of what changed in response (pushed in f2551d3):

- **Fix**: closing-tag-args misuse (rule 5) was silently lost for a single-`#` end tag carrying literal args — fixed with a separate `argsOnClose` token field, checked independently of `singleHash` so it doesn't hit the `tok.misuse` exclusivity trap you flagged.
- **Improvement** (literalBlockName drift): added `TestLiteralBlockNameMatchesV2BlockPatternArgClass` (new `internal_test.go`) rather than exporting the class from `codegen`, to avoid a public API change.
- **Suggestion** (named capture groups): done, in both `classify()` and `fixSingleHashLine`.
- **Improvement** (`EndBlock`-spelled / triple-`#` gaps): recorded as explicitly out of scope in the PR body — confirmed both are unreachable on `main` too, so this isn't a regression.
- **Documentation** (the `dynamic` derivation change): added a paragraph to "Notes for your reviewers" explaining it.

All four inline threads replied to and resolved individually. Re-requesting your review.

https://claude.ai/code/session_017FN9LymAqJ5DVX355PzYTt

---
_Generated by [Claude Code](https://claude.ai/code)_

marnagy (approved):

